PR #11, opened 2022-02-28
Cleanup higher-order-functions page

memowe:
The page appeared as duplicates in `docs/higher-order-functions.html` and `docs/higher-order-functions/index.html` with relative links being the only difference. I deleted the one in its own directory as all other chapters are just plain files in `docs`.

I updated all occurrences to the deleted page my editor could find, but please double-check.


smith558:
Thank you for this. I've noticed this weird redundant duplicate previously but didn't have time to clean it up. Thanks for your contribution!

memowe:
I'm glad to help :)

PR #17, opened 2022-03-08
Repair lines

memowe:
*Some* HTML files had *some* DOS line endings (CRLF instead of LF). I used
- `dos2unix docs/*.html` to convert line endings
- `sed -i 's/[ \t]*$//' docs/*.html` to remove trailing white spaces

Unfortunately the resulting diffs are unreadable. I suggest trying the same steps and comparing with my results before merging to make sure these commits don't break anything.


smith558:
@memowe would you mind looking at some of the merge conflicts with main which appeared on your branch? so we can merge this smoothly 🙂

memowe:
> @memowe would you mind looking at some of the merge conflicts

Of course. Instead of trying to merge I will pull the external changes and run my commands again as my diffs look horrible. ;)

memowe:
@smith558 Done. :) But please have a good look whether these commits break anything. I used exactly the commands documented here and in the commits.

smith558:
@memowe Thank you for your contribution! 🙂

PR #18, opened 2022-03-08
Cleanup duplicate style/reset css files

memowe:
The `style`/`reset` css files in `docs` and in `docs/assets/css` were the same. As most files reference the css files in `docs`, I removed the other ones.


memowe:
Rebased to latest main branch. :)

smith558:
Hi @memowe. Thanks for suggesting this change as well! 🙂 These duplicate CSS files have actually slipped in during some of my (in-progress) work on refactoring the source code. I'll need to check if this is not going to break something before merging. I should have been doing this on a separate branch and not on the main.

ulysses4ever:
this has been cleaned up as a part of #103, so I believe this can be safely closed.

PR #23, opened 2022-07-04
Mention complex numbers as an example why `Ord` is not a superclass of `Num`

konsumlamm:


smith558:
Thank you for your contribution! 🙂

PR #24, opened 2022-07-05
Update chapter 1 & 2

### LINKED ISSUE #16: Suggest GHCup instead of Haskell Platform
The [What you need to dive in](https://learnyouahaskell.github.io/introduction.html#what-you-need) section in the introduction currently recommends downloading the Haskell Platform. However, the Haskell Platform is [deprecated](https://www.haskell.org/platform/) and it is recommended to use [GHCup](https://www.haskell.org/ghcup/) instead, so the section should be updated.

konsumlamm:
* #haskell is now on Libera.Chat
* recommend GHCup as install method (fixes #16)
* replace "Haskell script" by "Haskell file"
* update GHCi output (I used version 8.10.7, since that's the version being recommended by GHCup)
* remove `let`s in GHCi that are now unnecessary
* fix typos


MatthijsBlom:
Thanks for doing this!

Note: [starting with GHC 9.0.1](https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/9.0.1-notes.html#ghci) the GHCi prompt will no longer list loaded modules, so this will have to be updated further once GHCup starts recommending GHC 9.2, though I see the old prompts further down are already consistent with future behaviour.

smith558:
This is indeed great. Thanks for your contribution! 🙂

PR #26, opened 2022-08-30
Rewrite guard example is Syntax in Functions to use Density (of substances) instead of BMI

### LINKED ISSUE #3: Replace Comments about BMI
The `syntax-in-functions` page includes a BMI example that may benefit from being rewritten. The two options I'm aware of would be:

* Keeping using BMI as the example, but change the messages for different categories
* Change the example to something else that uses division, perhaps calculating speed from distance and time, and then it would be possible to provide silly messages that aren't about a sensitive issue.

I prefer the second option because I think it would just be best to avoid discussing BMI. I am happy to put together a PR if the maintainer agrees.

dbp:
This fixes #3. 

That issue describes most of the motivation, but just to spell it out explicitly here: the BMI example is demonstrably offensive (plenty of examples, e.g., https://www.reddit.com/r/haskell/comments/sogi3s/learn_you_a_haskell_a_community_version/hwbmjg3/), and for good reason: not only is insulting people based on body size unpleasant, the example perpetuates really unhelpful stereotypes about high BMI people: first and foremost, that they cannot possibly be healthy (thus, if you have a BMI above a certain number, you need to lose weight). Nevermind that professional athletes regularly have "obese" level BMIs (as weight tells nothing about how your body actually functions), that weight has genetic factors, and that, in general, the type of "health" that people usually point to when making these kind of comments has a lot more to do with activity than about weight (which, despite assumptions, are not necessarily linked).

Regardless, this pull request keeps all of the same _technical_ content, with both a simple guard and then a guard that involves a calculation, and I attempted to include some of the light heartedness that characterizes the book.

### DIFF
--- a/docs/syntax-in-functions.html
+++ b/docs/syntax-in-functions.html
@@ -167,33 +167,33 @@ ghci&gt; capital "Dracula"
 <a name="guards-guards"></a><h2>Guards, guards!</h2>
 <img src="https://s3.amazonaws.com/lyah/guards.png" alt="guards" class="left" width="83" height="180">
 <p>Whereas patterns are a way of making sure a value conforms to some form and deconstructing it, guards are a way of testing whether some property of a value (or several of them) are true or false. That sounds a lot like an if statement and it's very similar. The thing is that guards are a lot more readable when you have several conditions and they play really nicely with patterns.</p>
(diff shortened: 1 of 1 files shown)


MatthijsBlom:
Looks good to me 👍

Thanks!

smith558:
Looks great. Thanks for the contribution! 🙂

dbp:
You’re welcome! And thanks for the great job with this fork!

ulysses4ever:
@dbp thanks for the contribution! BMI is still there in the [next section (`Where?!`)](https://learnyouahaskell.github.io/syntax-in-functions.html#where), though. Do you think you could fix it too?

dbp:
@ulysses4ever Thanks for noticing, and sorry! I've opened another pull request (#27) that should fix the rest, making it consistent with this one.

PR #29, opened 2022-09-19
Fixed strange wording

miestrode:
Whilst reading this part, I found the wording here to be both incorrect and cumbersome, unless I somehow misunderstood it. This PR changes it to fix that issue.


MatthijsBlom:
This looks good to me. Thanks!

It looks like this chapter needs updating: these days `catch :: Exception e => IO a -> (e -> IO a) -> IO a` instead of `catch :: IO a -> (IOError -> IO a) -> IO a`.

miestrode:
I honestly agree. As another example, the chapter is also outdated in the functions it uses for RNG. From what I know, one should use `uniform`, not `random` these days.

I might make a PR fixing this issue alongside others.

smith558:
> I honestly agree. As another example, the chapter is also outdated in the functions it uses for RNG. From what I know, one should use `uniform`, not `random` these days.
> 
> I might make a PR fixing this issue alongside others.

That'd be awesome if you do find time to open that PR!

MatthijsBlom:
By all means feel free to open issues for any problems you encounter, even if only to point them out to others.

smith558:
@miestrode Thank you for your contribution! 🙂

PR #34, opened 2022-10-03
Remove mistakenly added file diff_zippers.html

xogcox:
I just noticed that my earlier pull request adds the file `docs/diff_zippers.html`, which is just a renamed copy of the version of `zippers.html` prior to that PR. I was trying things out and added it by mistake. Since it doesn't serve any purpose, please delete it.

Sorry for the inconvenience.


smith558:
Not a problem. Thx for the corection!

PR #41, opened 2022-11-27
Make preparations for switching to markdown-generated HTML

xogcox:
In pull request #35 for issue #8, I added a markdown version of the website, but the HTML files generated from the markdown couldn't be used directly, because they required some changes to the CSS and JavaScript files. (Also, the relative links to assets were broken, because the generated HTML was stored in a different directory from the original HTML.)

The current pull request resolves this by changing the original website HTML step by step in a series of commits, ending up with HTML that is identical to the markdown-generated files. These commits include the necessary changes to CSS styles, and in one case change to a few lines of the JavaScript used to highlight code blocks. There are also a few changes to the markdown itself and the shell script generating the HTML. I broke the PR down into a series of separate commits in order to show what changes are required.

The result is a working website that is almost identical to the one before the changes, with only subtle differences: the only ones I am aware of are curved quotation marks replacing straight quotation marks in the main text and the trimming of blank lines at the end of a couple of code blocks.

Of course, merging this pull request does not mean you have to switch to generating the website from the markdown source, but it removes almost all of the obstacles to adopting markdown, if you decide to. It shows what needs to change in order for the markdown source to be actually used for building the website. Please ask if you have any questions or want more details.

PR #46, opened 2023-02-16
Use modern definitions for Monad and Alternative

### LINKED ISSUE #5: "The Monad type class" is outdated
The section "The Monad type class" in the chapter "A Fistful of Monads" (https://learnyouahaskell.github.io/a-fistful-of-monads.html#the-monad-type-class) talks about how `Applicative` is not a superclass of `Monad`. However, that is not true anymore. Additionally, `fail` is no longer a method of `Monad`.

Mr-Andersen:
Resolves #5

### DIFF
--- a/markdown/generated_html/a-fistful-of-monads.html
+++ b/markdown/generated_html/a-fistful-of-monads.html
@@ -254,29 +254,20 @@ check out the type class that belongs to monads.</p>
 applicative functors have the <code>Applicative</code> type class,
 monads come with their own type class: <code>Monad</code>! Wow, who
 would have thought? This is what the type class looks like:</p>
-<pre class="haskell:hs"><code>class Monad m where
+<pre class="haskell:hs"><code>class Applicative m =&gt; Monad m where
     return :: a -&gt; m a
+    return = pure
 
     (&gt;&gt;=) :: m a -&gt; (a -&gt; m b) -&gt; m b
 
     (&gt;&gt;) :: m a -&gt; m b -&gt; m b
-    x &gt;&gt; y = x &gt;&gt;= \_ -&gt; y
-
-    fail :: String -&gt; m a
-    fail msg = error msg</code></pre>
+    x &gt;&gt; y = x &gt;&gt;= \_ -&gt; y</code></pre>
 <p><img src="assets/images/a-fistful-of-monads/kid.png" class="right"
 width="363" height="451" alt="this is you on monads" /></p>
 <p>Let’s start with the first line. It says
-<code>class Monad m where</code>. But wait, didn’t we say that monads
-are just beefed up applicative functors? Shouldn’t there be a class
-constraint in there along the lines of
-<code>class (Applicative m) = &gt; Monad m where</code> so that a type
-has to be an applicative functor first before it can be made a monad?
-Well, there should, but when Haskell was made, it hadn’t occured to
-people that applicative functors are a good fit for Haskell so they
-weren’t in there. But rest assured, every monad is an applicative
-functor, even if the <code>Monad</code> class declaration doesn’t say
-so.</p>
+<code>class Applicative =&gt; Monad m where</code> which means that if
+we want to create an instance of Monad for some time, we must have an
+instance of Applicative for that type.</p>
 <p>The first function that the <code>Monad</code> type class defines is
 <code>return</code>. It’s the same as <code>pure</code>, only with a
 different name. Its type is <code>(Monad m) =&gt; a -&gt; m a</code>. It
@@ -305,22 +296,16 @@ returns a monadic value.</p>
 attention to it for now because it comes with a default implementation
 and we pretty much never implement it when making <code>Monad</code>
 instances.</p>
-<p>The final function of the <code>Monad</code> type class is
-<code>fail</code>. We never use it explicitly in our code. Instead, it’s
-used by Haskell to enable failure in a special syntactic construct for
-monads that we’ll meet later. We don’t need to concern ourselves with
-<code>fail</code> too much for now.</p>
 <p>Now that we know what the <code>Monad</code> type class looks like,
 let’s take a look at how <code>Maybe</code> is an instance of
 <code>Monad</code>!</p>
 <pre class="haskell:hs"><code>instance Monad Maybe where
-    return x = Just x
     Nothing &gt;&gt;= f = Nothing
-    Just x &gt;&gt;= f  = f x
-    fail _ = Nothing</code></pre>
-<p><code>return</code> is the same as <code>pure</code>, so that one’s a
-no-brainer. We do what we did in the <code>Applicative</code> type class
-and wrap it in a <code>Just</code>.</p>
+    Just x &gt;&gt;= f  = f x</code></pre>
+<p>Both <code>return</code> and <code>(&gt;&gt;)</code> have <em>default
+implementations</em>, so we omit them in instances. <code>return</code>
+is the same as <code>pure</code>, it wraps a value in
+<code>Just</code>.</p>
 <p>The <code>&gt;&gt;=</code> function is the same as our
 <code>applyMaybe</code>. When feeding the <code>Maybe a</code> to our
 function, we keep in mind the context and return a <code>Nothing</code>
@@ -811,17 +796,16 @@ matching in <code>let</code> expressions results in an error being
 produced right away, because the mechanism of falling through patterns
 isn’t present in <code>let</code> expressions. When pattern matching
 fails in a <code>do</code> expression, the <code>fail</code> function is
-called. It’s part of the <code>Monad</code> type class and it enables
-failed pattern matching to result in a failure in the context of the
-current monad instead of making our program crash. Its default
-implementation is this:</p>
-<pre class="haskell:hs"><code>fail :: (Monad m) =&gt; String -&gt; m a
-fail msg = error msg</code></pre>
-<p>So by default it does make our program crash, but monads that
-incorporate a context of possible failure (like <code>Maybe</code>)
-usually implement it on their own. For <code>Maybe</code>, its
+called. It’s part of the <code>MonadFail</code> type class and it
+enables failed pattern matching to result in a failure in the context of
+the current monad instead of making our program crash.</p>
+<pre class="haskell:hs"><code>class Monad m =&gt; MonadFail m where
+    fail :: String -&gt; m a</code></pre>
+<p>Monads that incorporate a context of possible failure (like
+<code>Maybe</code>) usually implement it. For <code>Maybe</code>, its
 implemented like so:</p>
-<pre class="haskell:hs"><code>fail _ = Nothing</code></pre>
+<pre class="haskell:hs"><code>instance MonadFail Maybe where
+    fail _ = Nothing</code></pre>
 <p>It ignores the error message and makes a <code>Nothing</code>. So
 when pattern matching fails in a <code>Maybe</code> value that’s written
 in <code>do</code> notation, the whole value results in a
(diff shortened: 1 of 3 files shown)


smith558:
I will wait for any more input on this and possibly a review before we merge. Thanks for the contribution! 🙂

MatthijsBlom:
It looks like you edited both `.md` and `.html` files. Is this intended?

Care should be taken that this PR does not introduce inconsistencies with the rest of the book. A priori I do not expect this problem to be localized to a single chapter.

Mr-Andersen:
@MatthijsBlom I ran `cd markdown; bash generate.sh` after editing MD files

MatthijsBlom:
So the `for-a-few-monads-more.html` edit is due to a faulty previous commit?

Mr-Andersen:
Yep

markdown/source_md/a-fistful-of-monads.md
```diff
@@ -205,26 +205,20 @@ Wow, who would have thought?
...
 
 ![this is you on monads](assets/images/a-fistful-of-monads/kid.png){.right width=363 height=451}
 
 Let's start with the first line.
-It says `class Monad m where`.
-But wait, didn't we say that monads are just beefed up applicative functors?
-Shouldn't there be a class constraint in there along the lines of `class (Applicative m) = > Monad m where` so that a type has to be an applicative functor first before it can be made a monad?
-Well, there should, but when Haskell was made, it hadn't occured to people that applicative functors are a good fit for Haskell so they weren't in there.
-But rest assured, every monad is an applicative functor, even if the `Monad` class declaration doesn't say so.
+It says `class Applicative => Monad m where` which means that if we want to create an instance of Monad for some time, we must have an instance of Applicative for that type.
```
konsumlamm:
```suggestion
It says `class Applicative => Monad m where`, which means that if we want to create an instance of `Monad` for some type, we must also have an instance of `Applicative` for that type.
```

markdown/source_md/a-fistful-of-monads.md
```diff
@@ -249,23 +243,16 @@ It's like function application, only instead of taking a normal value and feedin
...
 instance Monad Maybe where
-    return x = Just x
     Nothing >>= f = Nothing
     Just x >>= f  = f x
-    fail _ = Nothing
 ```
 
-`return` is the same as `pure`, so that one's a no-brainer.
-We do what we did in the `Applicative` type class and wrap it in a `Just`.
+Both `return` and `(>>)` have _default implementations_, so we omit them in instances.
```
konsumlamm:
```suggestion
Both `return` and `(>>)` have _default implementations_, so we can omit them in instances.
```

markdown/source_md/a-fistful-of-monads.md
```diff
@@ -779,19 +766,19 @@ When matching on a pattern in a function fails, the next pattern is matched.
 If the matching falls through all the patterns for a given function, an error is thrown and our program crashes.
 On the other hand, failed pattern matching in `let` expressions results in an error being produced right away, because the mechanism of falling through patterns isn't present in `let` expressions.
 When pattern matching fails in a `do` expression, the `fail` function is called.
-It's part of the `Monad` type class and it enables failed pattern matching to result in a failure in the context of the current monad instead of making our program crash.
-Its default implementation is this:
+It's part of the `MonadFail` type class and it enables failed pattern matching to result in a failure in the context of the current monad instead of making our program crash.
```
konsumlamm:
```suggestion
It's part of the `MonadFail` type class and it enables failed pattern matching to result in a failure in the context of the current monad, instead of making our program crash.
```

markdown/source_md/a-fistful-of-monads.md
```diff
@@ -962,33 +947,33 @@ ghci> [ x | x <- [1..50], '7' `elem` show x ]
 
 We apply `show` to `x` to turn our number into a string and then we check if the character `'7'` is part of that string.
 Pretty clever.
-To see how filtering in list comprehensions translates to the list monad, we have to check out the `guard` function and the `MonadPlus` type class.
-The `MonadPlus` type class is for monads that can also act as monoids.
+To see how filtering in list comprehensions translates to the list monad, we have to check out the `guard` function and the `Alternative` type class.
+The `Alternative` type class is for Applicatives that can also act as monoids.
```
konsumlamm:
```suggestion
The `Alternative` type class is for applicative functors that can also act as monoids.
```

markdown/source_md/a-fistful-of-monads.md
```diff
@@ -962,33 +947,33 @@ ghci> [ x | x <- [1..50], '7' `elem` show x ]
...
-    mzero = []
-    mplus = (++)
+instance Alternative [] where
+    empty = []
+    (<|>) = (++)
 ```
 
-For lists `mzero` represents a non-deterministic computation that has no results at all --- a failed computation.
-`mplus` joins two non-deterministic values into one.
+For lists `empty` represents a non-deterministic computation that has no results at all --- a failed computation.
```
konsumlamm:
```suggestion
For lists, `empty` represents a non-deterministic computation that has no results at all --- a failed computation.
```

konsumlamm (commented):
LGTM after these changes.

konsumlamm:
Afaik, the markdown version isn't used yet, so this change wouldn't be visible as it is rn.

cc @xogcox

xogcox:
Hi, thanks for pinging me!

It seems like I haven't fully communicated the status of the markdown files, and there's some confusion to clear up.

As you say, @konsumlamm, the markdown isn't in use yet. Not only that, nothing below the /markdown directory is in use yet. At the moment the markdown and generated html in this directory are a proof of concept, and there are lots of subtle differences from the live site, which is in the /docs directory. So all of the edits in this PR are edits to a work in progress that doesn't reflect the actual live website.

For clarity, maybe I should explain the course of events.

As you probably know, the original Learn You A Haskell website didn't have markdown, only raw HTML. So I wrote a markdown version that was as close as possible to the original version (while using HTML tags inside the markdown only when necessary). But as I said above, there are lots of little differences between the markdown-generated HTML and the original. Most of the differences can't be seen when viewing the website in the browser, but lots of tag names are different, and of course the line breaks aren't the same, so there are changes to almost every line.

Normally, if there's a markdown version available, it makes sense for the markdown to be the single source of truth. There's not much point editing the markdown if you have to edit the HTML by hand anyway. But I didn't feel like I could just say "Hey, delete all the original HTML and use my markdown version instead, it's much better!" Firstly, I don't have the authority to make that decision. Also, what if the HTML got replaced, and it broke the CSS layouts, for example? There could be new bugs that would be difficult to track down.

So I thought that, to clear the way to potentially adopting the markdown, I should document all the changes that would be required, step by step, to turn the current website's HTML into the HTML generated by the markdown, in the form of a commit history. That way there'd be less risk of sudden surprise bugs, and if there were any bugs or unintended differences, they could be traced back to the change that caused them more easily. It would also show exactly what changes would be caused by moving to the markdown version, which would hopefully reassure everyone that the change would be painless.

The result of this work is PR #41, which is quietly waiting to be looked at. It hasn't had much attention... maybe nobody was sure what its purpose was? I thought maybe people had lost interest or were busy with other things. But if you want to use the markdown, please take a look at that PR. It makes all the step-by-step changes necessary for the HTML in the /docs directory to match the markdown-generated HTML and also makes a few necessary edits to script files, CSS and Javascript so that they work properly with the HTML.

TL;DR I highly recommend that you consider pull request #41 before merging PRs that edit the markdown data.

It should be easier to merge #46 on top of #41 than the other way around. Also, after #41 you can just copy the generated HTML files into the /docs folder, since it makes the HTML files in /markdown/generated-html and /docs identical.

I'm glad to see that this project is active! Please feel free to give me feedback or ask if anything is unclear.

markdown/generated_html/a-fistful-of-monads.html
```diff
@@ -254,29 +254,20 @@ <h2 id="the-monad-type-class">The Monad type class</h2>
...
-constraint in there along the lines of
-<code>class (Applicative m) = &gt; Monad m where</code> so that a type
-has to be an applicative functor first before it can be made a monad?
-Well, there should, but when Haskell was made, it hadn’t occured to
-people that applicative functors are a good fit for Haskell so they
-weren’t in there. But rest assured, every monad is an applicative
-functor, even if the <code>Monad</code> class declaration doesn’t say
-so.</p>
+<code>class Applicative =&gt; Monad m where</code> which means that if
+we want to create an instance of Monad for some time, we must have an
```
xogcox: Should "for some time" be "for some type"?

markdown/generated_html/a-fistful-of-monads.html
```diff
@@ -305,22 +296,16 @@ <h2 id="the-monad-type-class">The Monad type class</h2>
...
-    return x = Just x
     Nothing &gt;&gt;= f = Nothing
-    Just x &gt;&gt;= f  = f x
-    fail _ = Nothing</code></pre>
-<p><code>return</code> is the same as <code>pure</code>, so that one’s a
-no-brainer. We do what we did in the <code>Applicative</code> type class
-and wrap it in a <code>Just</code>.</p>
+    Just x &gt;&gt;= f  = f x</code></pre>
+<p>Both <code>return</code> and <code>(&gt;&gt;)</code> have <em>default
+implementations</em>, so we omit them in instances. <code>return</code>
```
xogcox, line 306: Pandoc converts `_italic text markdown_` to `<em>` tags, but unfortunately, the current `/docs/style.css` file redefines `em` to be bold. That means that to include italic text, you might have to change the CSS.

markdown/generated_html/a-fistful-of-monads.html
```diff
@@ -254,29 +254,20 @@ <h2 id="the-monad-type-class">The Monad type class</h2>
...
-people that applicative functors are a good fit for Haskell so they
-weren’t in there. But rest assured, every monad is an applicative
-functor, even if the <code>Monad</code> class declaration doesn’t say
-so.</p>
+<code>class Applicative =&gt; Monad m where</code> which means that if
+we want to create an instance of Monad for some time, we must have an
+instance of Applicative for that type.</p>
 <p>The first function that the <code>Monad</code> type class defines is
 <code>return</code>. It’s the same as <code>pure</code>, only with a
 different name. Its type is <code>(Monad m) =&gt; a -&gt; m a</code>. It
```
xogcox, line 273: Because the line `return = pure` was added, maybe change this part to say not just that `return` always does the same as `pure` but that it's explicitly defined to be `pure`?

markdown/generated_html/a-fistful-of-monads.html
```diff
@@ -305,22 +296,16 @@ <h2 id="the-monad-type-class">The Monad type class</h2>
 attention to it for now because it comes with a default implementation
 and we pretty much never implement it when making <code>Monad</code>
 instances.</p>
-<p>The final function of the <code>Monad</code> type class is
```
xogcox, line 308: Maybe change the preceding paragraph to say that `>>` is the final function of the type class, instead of "next up", since the paragraph on `fail` has been deleted?

markdown/generated_html/a-fistful-of-monads.html
konsumlamm: Yes, this was already updated in the `.md` source, but the HTML hasn't been regenerated.

smith558:
@konsumlamm @xogcox would you be interested in getting collaborator rights? I see you are fairly active.

konsumlamm:
> @konsumlamm @xogcox would you be interested in getting collaborator rights? I see you are fairly active.

Sure! I already accepted the invitation.

xogcox:
Thank you for the invitation! I've joined too. I'll try to be useful.

smith558:
> Thank you for the invitation! I've joined too. I'll try to be useful.

No pressure. It's a fun project.

smith558:
Is this still ongoing?

PR #52, opened 2023-03-17
Change Monad and Alternative definitions, on the website

xogcox:
Since #46 only affects the `markdown` directory, it doesn't change the live website HTML, which is in the `docs` directory. This pull request applies the changes in #46 to the HTML in the `docs` directory as well.


smith558 (approved):

konsumlamm (changes requested):
There are some changes in #46, that haven't been included here (https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/46#pullrequestreview-1301926940). The html files weren't regenerated after these md changes, so they're outdated.

PR #55, opened 2023-08-30
remove extra word, regen html

maueroats:
Changed input-and-output.md, regenerated html
This resulted in changed paragraph flow in a few other html files. (Why?)


memowe (approved):
Thank you very much for your contribution <3 and sorry for the late reply.

PR #57, opened 2023-09-14
fix: Error in modules/transpose example

sravan-s:
Screenshot_20230914_163442


sravan-s:
Other guy fixed it too :)

PR #59, opened 2024-01-07
Add pattern guards

kephas:
Pattern guards are part of the guard syntax in `Haskell2010` and I think they should be included in the book. I'm not certain how to introduce them without making the existing text neither confusing nor too simplistic, but here's my first attempt.


smith558 (dismissed):
Looks good to me.

docs/syntax-in-functions.html
```diff
@@ -177,9 +177,9 @@ <h1 style="margin-left:-3px">Syntax in Functions</h1>
...
+<p>Guards are indicated by pipes that follow a function's name and its parameters. Usually, they're indented a bit to the right and lined up. A guard can be two things, the first is basically a boolean expression. If it evaluates to <span class="fixed">True</span>, then the corresponding function body is used. If it evaluates to <span class="fixed">False</span>, checking drops through to the next guard and so on. If we call this function with <span class="fixed">24.3</span>, it will first check if that's smaller than or equal to <span class="fixed">1.2</span>. Because it isn't, it falls through to the next guard. The check is carried out with the second guard and because <span class="fixed">24.3</span> is less than <span class="fixed">1000.0</span>, the second string is returned.</p>
```
ulysses4ever: I think this makes more sense:
```suggestion
<p>Guards are indicated by pipes that follow a function's name and its parameters. Usually, they're indented a bit to the right and lined up. A guard can be one of two things. The first is basically a boolean expression. If it evaluates to <span class="fixed">True</span>, then the corresponding function body is used. If it evaluates to <span class="fixed">False</span>, checking drops through to the next guard and so on. If we call this function with <span class="fixed">24.3</span>, it will first check if that's smaller than or equal to <span class="fixed">1.2</span>. Because it isn't, it falls through to the next guard. The check is carried out with the second guard and because <span class="fixed">24.3</span> is less than <span class="fixed">1000.0</span>, the second string is returned.</p>
```

kephas:
Thanks for the input @ulysses4ever, that's clearer that way!

smith558 (approved):

kephas:
@memowe @xogcox @konsumlamm is there something I should change in this PR?

ulysses4ever (approved):
LGTM, thanks!

smith558 (approved):

smith558:
@kephas Thank you for your contribution! 🙂

PR #71, opened 2024-09-04
fix a couple of typos

chad3814:
* on the homepage it says `be` instead of `by`
* on the 'Higher Order Functions' page it says
  > Let's answer us this question: How many elements does it take for the sum of the roots of all natural numbers to exceed 1000? To get the squares of all natural numbers, we just do map sqrt [1..]. 
  * That should be `To get the square roots of all natural numbers`


memowe (approved):
Thank you very much for your contribution! ❤️

PS: sorry for the delayed reply, I just got back from vacation.

smith558 (approved):

smith558:
Thank you for your contribution! 🙂 @chad3814

smith558:
Thanks for the review!! @memowe

chad3814:
thanks everyone, I'm just learning me a haskell…

PR #82, opened 2024-12-23
Minor refactoring of generate.sh

oliverkwebb:
I hope to get the build infrastructure in a flexible enough state eventually so that the book can be compiled into different formats (mainly, PDF and single-page HMTL)

Changes Made:
* Remove significant whitespace in generate.sh
* Give generate.sh executable permissions
* Re-run generate.sh, pushing a few previous changes to the generated html
* `sed command | sed command` -> `sed "command; command"`, unnecessary `$` in shell math  

P.S.
Right now, the filenames for the chapters are being stored in a simple text file, and the titles for the chapters are being extracted using `sed`, if that file was a YAML file, which stored the filenames and titles in a list, and the templating could use it to reference the previous/next chapters (with pandoc's `$for$`). It would remove the need for most of the shell code here (at the cost of the file being harder to parse)

### DIFF
--- a/markdown/generate.sh
+++ b/markdown/generate.sh
@@ -13,8 +13,8 @@ do
 
     title[$i]=$(sed -n '/^# /s/# //p;' $sourcemd | sed 's/{.*//; s/ *$//g')
 
-    chnum=$(($i + 1))
-    if [[ $chnum -ge 10 ]];
+    chnum=$((i + 1))
+    if ((chnum >= 10))
     then
         sp=" "
     else
@@ -30,7 +30,7 @@ done
 
 for i in ${!filename[@]}
 do
-    if (($i <= 0))
+    if ((i <= 0))
     then
         prev_title=
         prev_filename=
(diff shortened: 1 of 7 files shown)


ulysses4ever:
thanks for the contribution! this will be largely superceeded by #103. 

As for the `sed` business... Sadly, Pandoc alone will be rather unpleasant in handling prev/next links, because the easiest is to hardcode these relationships like this:

```
    ---
    items:
      - title: Item 1
        content: This is item 1.
        next_item:
          title: Item 2
          url: /item2.html
      - title: Item 2
        content: This is item 2.
        next_item:
          title: Item 3
          url: /item3.html
      - title: Item 3
        content: This is item 3.
        # No next_item for the last item
    ---
```

Of course, it's not too bad because we don't expect the links to change any time soon, and `sed` is extremely fragile and [non-portable](https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/103#discussion_r2606189620), so maybe it's not too bad of an idea. Of course, if we had something more pleasant to replace bash, like a Hakyll setup, that would be even better...

ulysses4ever (changes requested):
Hey @oliverkwebb! Could you please rebase on top of the current `main` branch? We don't store generated HTML anymore and the shell script  has been cleaned up a bit (including removing `$` in arithmetics) but some of your changes may still be relevant!

PR #88, opened 2025-06-18
wrong import for catch

JadAbouHawili:
catch is imported from Control.Exception , not System.IO.Error


docs/input-and-output.html
```diff
@@ -1199,16 +1199,14 @@ <h1>Input and Output</h1>
...
+<p>To deal with this by using exceptions, we're going to take advantage of the <span class="label function">catch</span> function from <span class="fixed">Control.Exception</span>. Its type is <span class="fixed">catch :: IO a -&gt; (IOError -&gt; IO a) -&gt; IO a</span>. It takes two parameters. The first one is an I/O action. For instance, it could be an I/O action that tries to open a file. The second one is the so-called handler. If the first I/O action passed to <span class="fixed">catch</span> throws an I/O exception, that exception gets passed to the handler, which then decides what to do. So the final result is an I/O action that will either act the same as the first parameter or it will do what the handler tells it if the first I/O action throws an exception.</p>
```
konsumlamm, line 1202: If we use `Control.Exception.catch`, the type needs to be changed to `Exception e => IO a -> (e -> IO a) -> IO a`. Alternatively, `System.IO.Error.catchIOError` could be used.

The Haskell report defines `System.IO.Error.catch :: IO a -> (IOError -> IO a) -> IO a` (which is where this issue comes from), which was then renamed to `catchIOError` in GHC (and MicroHs).

pierluc-codes:
Hey @JadAbouHawili  @konsumlamm, it looks like the code block are out of date. I am planning to revisit the whole book but if you have more suggestion in this chapter in the meantime, that would be amazing.

PR #95, opened 2025-11-19
removed propaganda and spyware

polymcwingetuploadguy:


smith558:
Hi @polymcwingetuploadguy !

Happy to remove Google Analytics, no use for them really (it's actually **disabled**, just the tag stayed from testing early on).

We'd like to keep Clicky, though; it's just to get an idea of how many people are visiting and which content to focus on. It's very useful in that way. You can read about why Clicky is not "spyware" here: https://clicky.com/help/faq/privacy. (but in short, it has **anonymised IP addresses** and **no tracking cookies**)

Can you clarify why a banner for UNICEF donations to Ukrainian children and a closable pop-up with information (attaching a screenshot below) for Russia-based visitors are "**propaganda**"? (I usually define it as “Propaganda” = deceptive, manipulative, or misleading political messaging, often state-controlled)

the pop-up:
image

the banner:
image

ulysses4ever:
Since the author didn't engage in the discussion, I propose we close this. But anyone should feel free to open a PR removing Google Analytics.

ulysses4ever:
Let's wait for one week to see if the author is still interested.

desjardinsalec:
I think this can be closed. Google Analytics has now been removed.

Describing the Ukraine donation as propaganda feels like a stretch, but I agree it should be removed. Since the maintainers disagree, this PR is no longer necessary.

pierluc-codes:
Thank you for the pull request.

The banner was a deliberate addition by a previous maintainer acting in good faith. We want to respect those editorial decisions, so I am closing this PR to keep that element of the site in place for now.

Regarding the technical side of the site, we have already moved toward a more privacy-focused approach by removing Google Analytics in a previous PR.

While we are keeping the current layout for the moment, we are open to re-evaluating the site's content if new information or a broader community consensus emerges. We’d prefer to handle those types of shifts through a dedicated Issue discussion first so we can keep the PR queue focused on technical Haskell documentation. Thanks for your interest in the project.

PR #92, opened 2025-10-14
Fix typos found by codespell

proski:
I haven't modified files with "generated" in the name, they should be regenerated. I don't know what's the right procedure for generation.

PR #67, opened 2024-07-01
wee -> we

bwkam:
^

PR #98, opened 2025-11-21
fix: generate.sh and delete generated files

pierluc-codes:
This PR generate.sh was not working because 

```yml
# file: markdown/config/pandoc-defaults.yml
highlight-style: null
```

By specifying the value, the script is not working.

I fix some typo, see comments inline.

@ulysses4ever and me decidd to delete the generated files (markdown and html) since it is very confusing.


markdown/config/template.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN"
    "https://www.w3.org/TR/html4/strict.dtd">
-<html lange="en">
```
pierluc-codes, line 3: Small typo

markdown/generate.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
pierluc-codes, line 1: Hardcoding the path of bash will force this execution plan. If the environment is using a different bash it won't work. This line will use the first one on the PATH.

markdown/generate.sh
```diff
@@ -1,4 +1,5 @@
...
+set -e
```
pierluc-codes, line 2: This will stop the script in case of failure.

ulysses4ever:
I plan to review this over the weekend. Thanks!

ulysses4ever:
Could you please split configuration changes from textial changes in separate commits? 

Fire textual changes, why do you only change generated files and not source files? I'd expect source files to be changed as well.

pierluc-codes:
@ulysses4ever 

Thanks for your feedback.

> Fire textual changes, why do you only change generated files and not source files? I'd expect source files to be changed as well.

There was modification to source that didn't update the generate files. I didn't change the source file, so it was 'catch up' for what is already on main :) For instance: https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/67

I would love to automate that at one point, but as a a first step, contributors should be able to run the generate script would be great :)

For instance, this contribution didn't generate the docs: https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/92

pierluc-codes:
@ulysses4ever I split the changes into two commits as requested :)

ulysses4ever (dismissed):
Makes sense, thank you! 

So, Markdown already has these changes... Do you know anything about the top-level docs directory: does it have it or not? A random change that I tried to check (a mention of Haskell Discourse in Introduction) doesn't seem to be included in the version inside the docs. So, we have three slightly different versions (at times, at least). This is not a great place to be in. There should be some rationale documented somewhere (e.g. readme), as well as what path we would like to take from 3 versions to one.

ulysses4ever:
After learning about the history of the `markdown` directory, I'm of the opinion that we should delete the `markdown/generated_html` directory, as it serves no purpose and only confuses people.

pierluc-codes:
@ulysses4ever I think we can also delete generated_markdown. I went ahead and deleted both.

The repo is configured to dismiss review on changes, so you will have to re approve.

ulysses4ever (dismissed):
Thanks! 

Could you please update the PR title and description with the summary of the actual changes?

pierluc-codes:
> Thanks!
> 
> Could you please update the PR title and description with the summary of the actual changes?

Done!

pierluc-codes:
@ulysses4ever can you re-approve when you get a chance? The repo is configured in a manner that review are auto dismiss on changes. Also, I do not want to bypass rules :smiley:

ulysses4ever (approved):

PR #102, opened 2025-11-25
Add new maintainers

smith558:


smith558:
> Very well! Maybe instead of emailing we should suggest opening an issue/ discussion?..

Yeah, was just thinking about it too.

PR #103, opened 2025-11-27
Switch to Markdown

ulysses4ever:
This is a new life for  #41 (opened in Nov 2022 and abandoned since) ~~and the project~~!
The PR's version of the site is currently deployed at  https://ulysses4ever.github.io/learnyouahaskell.github.io/.
Sadly, because of the many changes, both structural and little but pervasive, the HTML diff isn't very useful. (The Markdown diff is useful.) Hence, the reviewers are advised to open and eyeball the deployed version.

Below I detail main changes that this PR brings.

## Main Changes

[1] Makes HTML generated from `/markdown/source_md` straight to `/docs` with Pandoc (which becomes a dependency of the project).

- use the new Makefile to generate the HTML (`make`) and to clean it up (`make clean`).

[2] Removes all HTML files that can be generated: `docs/*.html` and `markdown/generated_html`, except `docs/index.html`. Also removes `markdown/generated_md` (only used as a temporary directory really). This was a HUGE point of confusion for external contributors --- having several versions of the same thing.

## How's HTML

[3] There are many small changes to the generated HTML vs the current one mostly because the current one is pretty irregular (no closing `</p>` tags, some weird combination of p and img tags, etc.). The many changes are meticulously recorded in the commit history inherited from #41.

Visually the result looks almost the same (**please, double-check!**), modulo slight changes in how text flows around images, but that doesn't seem consequential in any way.

## How's CSS

[4] There doesn't seem be any changes on `main` since #41 was forked except for the responsiveness PR #64, which I manually applied.

## How's textual content

[5] There has been many little text changes since #41 was forked. @pierluc-codes carefully re-implemented them here and I cross-chcecked it.

## Minor cleanups and updates

[6] GitHub Workflow will now deploy the website by generating HTML from MD (obviously).

[7] generate.sh was cleaned up a bit, including with `shellcheck`. 

- future TODO: create a CI workflow to check that shellcheck is clean.

[8] there was duplication of style.css/reset.css (they lived both under `/docs` and under `/docs/assets/css`). There's now only one version (`/docs/assets/css`).

[9] Similar to [8], JS files for the syntax highlighter that lived under `/docs` were moved to a more appropriate place under `/docs/sh` (the directory was there before and held one file already).

[10] The new `.gitignore` excludes the generated files under `/docs`.

---

_old description_ 

TODO:

- [x] add textual changes since 2022; they have been collected by @pierluc-codes here: https://gist.github.com/pierluc-codes/9c07c421c087ab114c4456f6eb4f44ae
  - the current idea: @ulysses4ever goes from the top of the list in the Gist, and @pierluc-codes goes from the bottom. But we are both busy, so any help is welcome.

- [x] merge CSS: #41 had some changes, and the upstream had some, especially in #64; this includes:
  - [x] make sure Markdown-produced HTML fits with the CSS updates upstream
  - [x] spot-check that the visual changes aren't too big; some examples here: https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/41#issuecomment-3573248287
  - [x] nuke old html and possibly other generated files. This is not completely trivial because the generate.sh script starts failing of you simple remove that stuff. I need to understand better how the script works before I can act on this item.
  - [x] de-dup CSS files (eg style.css)

---

The current version is rendered here: https://ulysses4ever.github.io/learnyouahaskell.github.io/chapters.html Note for @pierluc-codes : I had to change the repository I worked from (it used to be ulysses4ever/learnyouahaskell-md) because I cant open pull requests from non-forks.

### DIFF
--- a/docs/assets/css/style.css
+++ b/docs/assets/css/style.css
@@ -130,7 +130,6 @@ pre > code {
     font-weight:bold;
     /* Allow inline code to wrap if necessary */
     /* These rules were based off the CSS used on the Stack Exchange website */
-    white-space-collapse:preserve;
     overflow-wrap:break-word;
     text-wrap:wrap;
     /* Rounded corners help user connect two parts of one code snippet wrapped across lines */


pierluc-codes:
@ulysses4ever First thank a lot for adding the GitHub workflow, it helped a lot in this instance.

I brought all the textual change. There were few instance where the textual change was wrong so I took the liberty to update it. Feel free to adjust.

Let me know if you need more help!

ulysses4ever:
@pierluc-codes terrific, thank you! (and sorry I wasn't around this week: stuff happens...).

I'll try to review the textual changes by the end of the week. I'm pretty sure, they should be fine!

In the meantime, we still should make sure styles are reasonable. [In this comment](https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/41#issuecomment-3573248287) you noticed several things, which aren't too bad, but the way the error message is formatted, probably needs some thought. Do you want to try to work on this front more? I see two options: (1) try to eyeball the text more and make a list of all noticeable styling changes; (2) try to fix some of these changes by updating CSS / MD / the way HTML is produced so that we have a better match with the original. THe second one is obviously more involved, so it's okay to take the first one or none at all. I hope to get to this by the end of week as well...

pierluc-codes:
@ulysses4ever I eyeballed the doc and I think we are good. I did some correction here and there.

There is still a minor difference related to image size but to be honest I like the bigger images ;)

I would like to note that migration also fixed few mistake that were in the original HTML. Very cool.

ulysses4ever:
Terrific! There are two big items on the list left (at least):

- integrate CSS changes from https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/64 so that we have a decent mobile view 

- nuke old html and the generated files. This is not completely trivial because the generate.sh script starts failing if you simply remove that stuff. I need to understand better how the script works before I can act on this item.

pierluc-codes:
@ulysses4ever I will be busy this week so I won't be able to push this forward for at least the next week. I can make time if you need a review on something.

ulysses4ever:
@pierluc-codes sounds good. I'll do my best!

ulysses4ever:
Just an update: I managed to remove dependence on generated files and removed the files. I switched to responsive CSS but got stuck because there are two slightly different style.css in the repo, and they have existed from the very beginning. I want to kill one.

ulysses4ever:
The responsive CSS patch from #64 is applied now. Next: merge upstream main to here so that we don't see all the conflicts.

ulysses4ever:
I intended to do more spot-checking but this PR is ready for review now. @pierluc-codes

markdown/config/pandoc-defaults.yml
```diff
@@ -1,3 +1,4 @@
 from: markdown-implicit_figures
 to: html
+highlight-style: null
```
ulysses4ever: @pierluc-codes this was in the original PR and you removed it recently. I'm not sure what's going on here: I haven't studied this Pasndoc option. My guess would be that we make sure Pandoc doesn't try to highlight anything because we supply our own JS highlighter. The result seem fine with this option at least?

ulysses4ever:
I made a big final sweep and this is ready for review now. This is a big change, so I'll ping all codeowners hoping to get more eyes on it. The result would benefit from eyeballing more. I also update this PR's description (see at the top) with a list of notable changes. Please, check it out...

It'd be good to merge it this week because there are many PR's on the repo that will benefit from rebasing on this.

@smith558
@konsumlamm
@xogcox
@memowe
@pierluc-codes

markdown/config/pandoc-defaults.yml
pierluc-codes: Interesting, on one of my mac (pandoc 3.8) the options does not work, on my linux (pandoc 3.1) machine it does work. One what are you working? Does the `null` works for you?

ulysses4ever: I'm on Linux and as far as I can see it works for me. Can you show me a screenshot of how exactly it looks when not working properly? And also, can you try the matching versions of pandoc? (I'd expect the versions to be more relevant than platforms in this issue.) Lastly, can you try different browsers on the Mac?

pierluc-codes: Screenshot 2025-12-09 at 06 02 29

ulysses4ever: Oh, interesting, I'll have to think about that... 

Did you notice that the script fails even earlier: it can't find `mapfile`. I'm guessing the results on Mac are currently invalid... We could describe this situation in readme and suggest Docker maybe.

pierluc-codes: Oops, I was running the old version of the script. There is also a problem with the first line of the script which force an old bash to run on osx. mapfile is not available on this version.

This what it looks like witth the new bash

Screenshot 2025-12-09 at 06 27 32

That is why I also added `set -e` to have earlier failure.

Anyway, it is beside the point. I think we can agree to migrate to a container in the nearby future to remove all those obstacles :).

ulysses4ever: Reading Pandoc help, I'm not even sure null is a valid value for this field. I guess we can remove this line (again). And open an issue able mac compatibility.

ulysses4ever: I removed the option. When you have a moment, please, try it on your Mac: if it finishes clean, we can simply let it go.

pierluc-codes: There is some incompatibility both BSD sed and GNU sed

```
sed: 1: "/<p><img/ { N; N; N; s# ...": bad flag in substitute command: '}'
make: *** [site] Error 1
```

I was able to run it after install GNU sed

```
brew install gnu-sed
export PATH="/usr/local/opt/gnu-sed/libexec/gnubin:$PATH"
make site
```

ulysses4ever: Good, thank you for looking into it. We should document it at some point but that's not on the critical path, unlike the Markdown overhaul. 

The bits of sed are scary, and I hope to migrate away from it in the future. FTR, this has been brought up before, e. g. https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/82

ulysses4ever:
Hey all, it was a quiet week!.. @smith558 do you think you could conduct a formal review any time soon? It doesn't have to be in-depth, just the general approval for the direction of travel would be great (and we do require one approval at least...).

@pierluc-codes any plans to give it a final review? I hope we can merge it before New Year!

pierluc-codes:
@ulysses4ever I did two minor correction. I think we are good. Feel free to merge when you are comfortable to.

Issue #104, opened 2025-12-03
Typo in Files and Streams


jaye311:
**Is your content edit request related to a problem? Please describe.**
I think there is a typo in describing piping to output when it should be piping to input.
"We pipe the contents of shortlines.txt into the **output** of shortlinesonly and as the output, we only get the short lines.

This pattern of getting some string from the input, transforming it with a function and then outputting that is so common that there exists a function which makes that even easier, called interact. interact takes a function of type String -> String as a parameter and returns an I/O action that will take some input, run that function on it and then print out the function's result. Let's modify our program to use that.

"
**Describe the solution you'd like**
A typo correction would be nice if this was a typo.

**Describe alternatives you've considered**
Maybe it's not a typo?

**Additional context**
<img width="1056" height="431" alt="Image" src="https://github.com/user-attachments/assets/282e95fd-e2f1-4f2b-b607-d3ca93c01df5" />

**Link to the page in question**
https://learnyouahaskell.github.io/input-and-output.html#files-and-streams

ulysses4ever:
Hey! Thanks for the report! You're absolutely right. Do you want to submit a PR fixing this? Note that only the contents of the `docs` directory are currently relevant.

desjardinsalec:
@ulysses4ever we can close based on #106 😄

PR #106, opened 2025-12-26
:recycle: Clarify pipe usage in shortlines example

desjardinsalec:
Fix for https://github.com/learnyouahaskell/learnyouahaskell.github.io/issues/104

Rather than swapping output for input, I _attempted_ to make the sentence more expressive/clear.

### DIFF
--- a/markdown/source_md/input-and-output.md
+++ b/markdown/source_md/input-and-output.md
@@ -788,7 +788,7 @@ so am i
 short
 ```
 
-We pipe the contents of *shortlines.txt* into the output of *shortlinesonly* and as the output, we only get the short lines.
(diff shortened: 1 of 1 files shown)


markdown/source_md/input-and-output.md
```diff
@@ -788,7 +788,7 @@ so am i
 short
 ```
 
-We pipe the contents of *shortlines.txt* into the output of *shortlinesonly* and as the output, we only get the short lines.
+We pipe the contents of *shortlines.txt* into *shortlinesonly*, and the output contains only the short lines.
```
pierluc-codes, line 791: Nice! Good catch!

PR #110, opened 2025-12-28
Move from Bash/sed to Haskell/Hakyll for site generation

ulysses4ever:
Hey all and happy holidays! I've spent my holidays (in part) porting the shell script (`generate.sh`) that converts Markdown to HTML to Haskell using the Hakyll library that assists in producing static websites.

I submit two commits: one just moves stuff from the `markdown` directory to the top level (makes sense after markdown became the only way to work with the website) and the other does the thing; this should ease reading the diff. In sections below, I discuss some high-level questions because I don't expect people to dive too deep in the Haskell implementation. Also, eyeballing the result is very welcome (see https://ulysses4ever.github.io/learnyouahaskell.github.io/). 

Those who are interested in the implementation, though, can find it in the single Haskell file (sans `templates`), [`site.hs`][sitehs], which I tried to document with comments as much as possible. The resulting script has grown about 1.5x in size relative to `generate.sh`, but I think it's more readable, and you can follow it even if you're not familiar with Hakyll. The code is co-authored by GitHub Copilot.

[sitehs]: https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/110/changes#diff-0b91bb1a7a517ebdc504c0350f6f8e217d964d075ffdc771c54c14fe9b0b6af9

## Why move away from the shell script

1. Hard to maintain (subjectively). It uses pretty gnarly `sed` expressions to generate the table of contents (`chapters.html`), for instance.

2. It's not entirely portable (e.g. MacOS users reported some troubles, although they can be resolved).

## Why use Haskell/Hakyll

1. First in class conversion tools: Pandoc, Hakyll's high-level DSL to describe routes and transformations, templating system

2. More portable. 

3. It makes a point to use Haskell to produce a book about Haskell. 

## Downsides of using Haskell 

1. Haskell is less known than Bash among the general population, so contributions and maintenance may be trickier. But there are challenges with bash/sed as well, as noted above. Overall, I think we are in profit. Also, casual contributors don't have to build the site: in most cases, editing markdown should be enough.

2. Bash/sed are omnipresent (the other side of the portability issue of course), whereas to build a Haskell application, you need special tools (I explain it in detail in the README). Again, just like with (1), I believe this downside is outweighed by the benefits.

## Changes in HTML

We still use Pandoc, so the results are mostly intact. There are two simplifications I made to make the whole thing simpler.

1. Small margins for some H1 headers are gone. There were (seemingly) arbitrary margins for the chapter headings. They looked like artifacts from the past (e.g. WYSIWYG tools sometimes produce this sort of small random adjustments). So, I removed them.

2. Img-tags are left wrapped in p-tags. Generate.sh patched those with sed, and I could certainly re-implement it in Haskell but I see no point in this: visually, the result seems the same.

I believe these changes don't have any negative effect while simplifying the whole picture.


source_md/chapters_head.md
ulysses4ever, line 1: This is a leftover that I should kick out.

README.md
```diff
@@ -38,6 +38,70 @@ Discuss the ideas with the community.
...
+## Contributing
+
+We are happy to get your contributions!
+For the most part, you can simply edit Markdown files and open a PR with the edits.
+If you want to preview the changes locally, you need to build the site as discussed below.
+
+### Building the site
+
+The site is built using [Hakyll](https://jaspervdj.be/hakyll/), a static site generator in the form of a Haskell library.
+With Hakyll, you first build a Haskell application `site`, and then run it to generate HTML for the website.
```
ulysses4ever: This should make a note that the two steps can be combined in some workflows (e. g. `cabal run`).

lyah-site.cabal
```diff
@@ -0,0 +1,21 @@
+cabal-version:      3.0
+name:               lyah-site
+version:            0.1.0.0
+license:            BSD-3-Clause
+author:             LYAH Community
+maintainer:         lyah-community@example.com
+build-type:         Simple
+tested-with:        GHC ==9.10
```
ulysses4ever: Should be 9.10.3.

lyah-site.cabal
```diff
@@ -0,0 +1,21 @@
...
+  main-is:          site.hs
+  build-depends:    base >= 4 && < 5
+                  , hakyll >= 4.16 && < 4.17
+                  , pandoc
+                  , pandoc-types
+                  , filepath
+                  , containers
+                  , text
+                  , directory
+  ghc-options:      -threaded
```
ulysses4ever: Perhaps, -Wall -Werror would be nice

site.hs
```diff
@@ -0,0 +1,206 @@
...
+{-
+
+  The general idea of a Hakyll website is trivial: convert markdown files to HTML (using Pandoc under the hood)
+  and put the results, as well as the static assets, in the output directory, `_site`.
+  However, there are two non-trivial tasks with our site:
+
+  1. Generating `chapters.html`, which is a table of contents (TOC) for all chapters, including subsections.
+  2. Back and forward links between chapters, so we need to know the order of chapters.
+
+  To solve both, we first use Pandoc to parse all chapter markdown files first to extract their chapter numbers and titles
```
ulysses4ever: And a list of sections!

site.hs
```diff
@@ -0,0 +1,206 @@
...
+            Right pandocDoc@(Pandoc meta blocks) ->
+                -- If there's no `chapter` field, it's not a chapter file (e.g., FAQ), and we return Nothing
+                M.lookup "chapter" (unMeta meta) <&>
+                    \chapterMeta ->
+                        ChapterInfo
+                            { chapterFile = fname
+                            , chapterNumber = read . T.unpack $ stringify chapterMeta
+                            -- RE fromJust below: every chapter is ought to have a title field
+                            , chapterTitle = T.unpack . stringify . fromJust $ M.lookup "title" (unMeta meta)
+                            , chapterSections = extractTOCFromPandoc pandocDoc
```
ulysses4ever: Pass just blocks

pierluc-codes:
Impressive work @ulysses4ever! :clap: 

I don't think I will have time to review before new year, but I will make time before the 5th for sure.

Looking forward to approve this! :)

ulysses4ever:
@pierluc-codes no rush! Thanks for replying and happy holidays!

pierluc-codes:
@ulysses4ever My apologies. Coming back from the holiday was much more difficult than anticipated. I will do a first round of review this weekend.

ulysses4ever:
@pierluc-codes no worries! Thanks for the update and looking forward to it!

templates/chapters.html
```diff
@@ -0,0 +1,15 @@
+<ol class="chapters" type="1">
+$for(chapters)$
```
pierluc-codes, line 2: Would it be possible to add tabs here? :)

.gitignore
```diff
@@ -1,4 +1,11 @@
-docs/*.html
```
pierluc-codes, line 1: I think we should leave those gitignore for now because people probably have them on their computer. Good idea to add those though

README.md
```diff
@@ -38,6 +38,70 @@ Discuss the ideas with the community.
...
+
+# Build and run the site generator
+cabal run site -- build
+
+# Preview the site locally (optional)
+cabal run site -- watch
+# Then visit http://localhost:8000
+```
+
+A variation of these commands is used in the `Makefile`: run `make` to generate the website.
```
pierluc-codes, line 73: Maybe consolidate?

README.md
```diff
@@ -38,6 +38,70 @@ Discuss the ideas with the community.
...
+
+#### Option 2: Using Nix
```
pierluc-codes, line 75: I didnt test the nix path, i am assuming it work fine since it is used by the CI

lyah-site.cabal
```diff
@@ -0,0 +1,21 @@
+cabal-version:      3.0
+name:               lyah-site
+version:            0.1.0.0
+license:            BSD-3-Clause
```
pierluc-codes, line 4: I dont think any this sorrect

lyah-site.cabal
```diff
@@ -0,0 +1,21 @@
...
+author:             LYAH Community
+maintainer:         lyah-community@example.com
```
pierluc-codes: there few field that are incorrect here

ulysses4ever:
@pierluc-codes thank you for the comments! I'm having an exceptionally busy January so I can attend to this only in a week at the earliest (maybe later). If you have cycles to fix simple things, please, feel free to commit into my branch here.

lyah-site.cabal
smith558, line 4: Licence should be `CC-BY-NC-SA-3.0`

ulysses4ever: But this is not the text, this is a Haskell app that I wrote (with Copilot). Why would it need to have the same license as the text @smith558 ?

As a meta-reason: picking the license you point to makes nix unhappy because it considers it a non-free licence. This can be worked around but requires setting environment variables which needs documentation and all, and I didn't want to create another obstacle for contributors. I don't remember for sure but cabal check might be unhappy about this license too. I'd prefer to publish code under a free license, which may save us some trouble down the road.

ulysses4ever: To be clear, this is not a hill I'll die on. So if everyone thinks that it's critical to have code under the same license, I'm happy to follow. I just want to understand the reasoning for it.

ulysses4ever: What would you prefer @pierluc-codes ? I don't see an issue with these fields.

README.md
ulysses4ever: What do you mean?

lyah-site.cabal
ulysses4ever: I removed the maintainer field (with the fake email), but I left "LYAH Community" as the author in place  because it is a good option for Haskell code written to maintain the community edition of the book, I feel. Let me know if you disagree @pierluc-codes.

pierluc-codes: Thanks! All good except the license that I am not totally sure about. I will let the other thread resolve it :)

README.md
pierluc-codes: I was thinking maybe just document one of the variant (Makefile or cabal directly).. or maybe next to each other.

ulysses4ever: I don't see how it's better than the current version or, indeed, what's the issue with the current version that you're trying to address with this suggestion. If you could propose an exact wording that you think would be an improvement, I'm happy to discuss.

pierluc-codes: I am not totally sure to get why the documentation reflect cabal when we have makefile goal.

Personally, I would have documented the makefile variant, not directly invoked cabal since the makefile is a little bit more complete.

Not a hill I will die on, but I like to remove confusion by giving an "correct" way to do it (which I think makefile would be more interesting).

ulysses4ever: Oh, I see. Yes, I can put Makefile forward. Thanks for elaborating!

lyah-site.cabal
smith558: Hi @ulysses4ever, sorry, this is my bad. `BSD-3-Clause` is completely fine here. I misunderstood what this is. I initially thought this was the text content metadata to be injected into the exported output by the tool.

ulysses4ever: Got you, thanks! No, it's just a glorified version of that bash script that currently runs Pandoc (the Haskell script also runs Pandoc under the hood). The Haskell script was implemented from scratch.

README.md
ulysses4ever: @pierluc-codes I started doing this and remembered one thing: we cannot completely hide `cabal` behind `make` for a simple reason: `make` will fail miserably if cabal/ghc aren't installed. So, I can show make-based commands first and mention that one may use cabal directly, but I'll have to start from explaining the dependency on the Haskell toolchain (ghc/cabal) still.

pierluc-codes: Got it. Let leave it that way for now.

ulysses4ever: Sure, thank you!

templates/chapters.html
ulysses4ever: I have added a version with tabs. Could you check? (layout is subjective, so I'm not sure what would make you happier)

.gitignore
ulysses4ever: I first thought it's a good idea but after thinking a bit more I'm not sure. `docs` won't exist after this is merged. If people have local `docs`, it'd be in their best interest to see that these aren't tracked anymore and the results go elsewhere, and they should probably get rid of the dead wood asap to avoid confusion. This is the right message. Instead, if it stays in gitignore, they might miss the change. 

Am I missing something?

pierluc-codes: ok

ulysses4ever:
@smith558 if you're happy with it, could you remove your request for changes / approve, please?

ulysses4ever:
@smith558 could you, please, re-review or, at least, remove request for changes. Otherwise I'm stuck. (I don't want to override unless you explicitly suggest that.)

ulysses4ever:
I went ahead and pulled the trigger since this has been lingering for too long. I'm happy to revert if anyone feels like it.

Issue #113, opened 2025-12-31
Input and Output - Files and Stream paragraph clean up


desjardinsalec:
**Is your content edit request related to a problem? Please describe.**
Update sentence describing cat palindrome text file into cat words.txt | runhaskell palindromes.hs.

> Again, we get the same output as if we had run our program and put in the words ourselves at the standard input. We just don’t see the input that palindromes.hs because the input came from the file and not from us typing the words in.

**Describe the solution you'd like**
The sentence below the code snippet should be clearer. 

**Describe alternatives you've considered**
An alternative:

> Again, we get the same output as if we had run the program and typed the words ourselves at the standard input. The difference is that we don’t see the input, because it was piped into palindromes.hs from a file rather than entered directly at the keyboard.

**Link to the page in question**
https://github.com/learnyouahaskell/learnyouahaskell.github.io/blob/main/markdown/source_md/input-and-output.md

<img width="1166" height="290" alt="Image" src="https://github.com/user-attachments/assets/c294028e-b124-4a2e-99e1-de58ee0ef7ef" />

PR #116, opened 2026-01-01
Fix formatting and reword paragraph for clarity

desjardinsalec:
https://github.com/learnyouahaskell/learnyouahaskell.github.io/issues/113

- Fix some minor formatting in one of the code snippets
- Reword paragraph slightly for clarity

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 Hey yo! [learnyouahaskell.github.io](https://learnyouahaskell.github.io/) is an **open-source** fork of the original **Learn You a Haskell** (**LYAH** for short) guide for Haskell by Miran Lipovača, "the funkiest way to learn Haskell, the best functional programming language around". This guide is meant for people who have programmed already, but have yet to try functional programming.
 
 **Now, why did I create this fork?** ("fork" means a copy of an original which may be modified or extended)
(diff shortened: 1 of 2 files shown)


ulysses4ever:
Thanks! This mostly looks great. One thing: I'm not a big fan of using the verb "to pipe" as it's a bit jargony, and not everyone will be familiar with this jargon.

desjardinsalec:
@ulysses4ever updated slightly, used provided.

markdown/source_md/input-and-output.md
```diff
@@ -882,8 +882,7 @@ palindrome
 palindrome
 ```
 
-Again, we get the same output as if we had run our program and put in the words ourselves at the standard input.
-We just don't see the input that `palindromes.hs` because the input came from the file and not from us typing the words in.
+Again, we get the same output as if we had run the program and typed the words ourselves at standard input. The difference is that we don’t see the input, because it was provided to `palindromes.hs` from a file rather than entered directly at the keyboard.
```
pierluc-codes: Sorry to bike shred, can you put each sentence on their own (markdown) line? It should not change the output, and it is easier to compare the diff in the future :)

ulysses4ever: One sentence per line is the approach I like the most too.

desjardinsalec: Will do, is the copy acceptable?

desjardinsalec: @ulysses4ever @pierluc-codes resolved. I will make sure to stick to that format.

ulysses4ever (approved):

PR #121, opened 2026-01-05
Correcting the generated smugpig.pngs height declaration.

chrisdenman:
The height 186 is incorrect, 168 is the correct value.

### DIFF
--- a/markdown/source_md/a-fistful-of-monads.md
+++ b/markdown/source_md/a-fistful-of-monads.md
@@ -5,7 +5,7 @@ Then, we took that concept one step further by introducing applicative functors,
 
 In this chapter, we'll learn about monads, which are just beefed up applicative functors, much like applicative functors are only beefed up functors.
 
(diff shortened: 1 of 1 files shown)


pierluc-codes:
@chrisdenman Thanks for your contribution! I'm not against it, I would like to undertand. Can you explain where the value is coming from? :)

chrisdenman:
Please see the attached pictures:

One shows the pig image and its actual dimensions. Note the height of 168px.

The other shows the source file `markdown/source_md/a-fistful-of-monads.md` which includes the image with this line `more cool than u{.right width=307 height=186}`. Note the height of 186px.

This causes the browser to pre-layout the page with an incorrect height value. Whilst the browswer locates the image, determines the actual size, the page may jump. 

Again, it's not a big deal but generates a warning in my IDE and an error in my functional test case (I ensure all images declare their actual sizes).

Thanks.

Screenshot from 2026-01-06 16-28-46

Screenshot from 2026-01-06 16-29-35

PR #133, opened 2026-02-02
Update syntax-in-functions.md

ellyxir:
italicise "where" since it's a keyword

### DIFF
--- a/source_md/functors-applicative-functors-and-monoids.md
+++ b/source_md/functors-applicative-functors-and-monoids.md
@@ -50,7 +50,7 @@ We saw how we can map functions over them for great good.
 In this section, we'll take a look at two more instances of functor, namely `IO` and `(->) r`.
 
 If some value has a type of, say, `IO String`, that means that it's an I/O action that, when performed, will go out into the real world and get some string for us, which it will yield as a result.
-We can use `<-` in *do* syntax to bind that result to a name.
+We can use `<-` in `do` syntax to bind that result to a name.
 We mentioned that I/O actions are like boxes with little feet that go out and fetch some value from the outside world for us.
 We can inspect what they fetched, but after inspecting, we have to wrap the value back in `IO`.
 By thinking about this box with little feet analogy, we can see how `IO` acts like a functor.
@@ -65,11 +65,11 @@ instance Functor IO where
         return (f result)
 ```
 
-The result of mapping something over an I/O action will be an I/O action, so right off the bat we use *do* syntax to glue two actions and make a new one.
+The result of mapping something over an I/O action will be an I/O action, so right off the bat we use `do` syntax to glue two actions and make a new one.
 In the implementation for `fmap`, we make a new I/O action that first performs the original I/O action and calls its result `result`.
 Then, we do `return (f result)`.
 `return` is, as you know, a function that makes an I/O action that doesn't do anything but only presents something as its result.
-The action that a *do* block produces will always have the result value of its last action.
+The action that a `do` block produces will always have the result value of its last action.
 That's why we use return to make an I/O action that doesn't really do anything, it just presents `f result` as the result of the new I/O action.
 
 We can play around with it to gain some intuition.
@@ -754,8 +754,8 @@ Since `pure` is all about putting a value in a minimal context that still holds 
 
 If `<*>` were specialized for `IO` it would have a type of `(<*>) :: IO (a -> b) -> IO a -> IO b`.
 It would take an I/O action that yields a function as its result and another I/O action and create a new I/O action from those two that, when performed, first performs the first one to get the function and then performs the second one to get the value and then it would yield that function applied to the value as its result.
-We used *do* syntax to implement it here.
-Remember, *do* syntax is about taking several I/O actions and gluing them into one, which is exactly what we do here.
+We used `do` syntax to implement it here.
+Remember, `do` syntax is about taking several I/O actions and gluing them into one, which is exactly what we do here.
 
 With `Maybe` and `[]`, we could think of `<*>` as simply extracting a function from its left parameter and then sort of applying it over the right one.
 With `IO`, extracting is still in the game, but now we also have a notion of *sequencing*, because we're taking two I/O actions and we're sequencing, or gluing, them into one.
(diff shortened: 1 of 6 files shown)


ulysses4ever:
The styling of this particular element seems to be all over the place (I only looked through that chapter), so I'm not sure how helpful it is to fix just one place. I'd probably prefer not using italics in phrases like "where blocks" at all. I'd also prefer consistently using hyphens, because I think "more" grammatically correct would be to say "where-blocks", but that's not strictly related to the italics, you may think, although "_where_-blocks", so italics for a part of a composite, would look weird.

You say "italicise "where" **since it's a keyword**": I'm not sure using italics for keywords is very popular in general. A monospaced font (or whatever is used for marking up inline code) would be more common.

ulysses4ever:
I guess if you could update the PR to fix all places like that, that would be an improvement on the current state, and something I'm happy to approve.

Thanks a lot for taking the time to submit the fix though!

smith558:
Agree with @ulysses4ever, whichever way it goes (italics, monospace, or no distinct "keyword style" whatsoever), consistency (both within and across all the chapters) is super important for the reading and learning experience.

ellyxir:
> The styling of this particular element seems to be all over the place (I only looked through that chapter), so I'm not sure how helpful it is to fix just one place. I'd probably prefer not using italics in phrases like "where blocks" at all. I'd also prefer consistently using hyphens, because I think "more" grammatically correct would be to say "where-blocks", but that's not strictly related to the italics, you may think, although "_where_-blocks", so italics for a part of a composite, would look weird.
> 
> You say "italicise "where" **since it's a keyword**": I'm not sure using italics for keywords is very popular in general. A monospaced font (or whatever is used for marking up inline code) would be more common.

ah, im going through the book now and i noticed it was italicised earlier so when i read this particular one, i had a moment of parsing where as the word where in english and not the keyword which confused me. not a big deal but i thought it was consistently italicised. i can take a look through the other uses of it.

pierluc-codes:
I think in this specific instance, i think we shoud use "where-blocks" (hyphenated). 

@ellyxir did you have a chance to tackle the other places? I can take a look otherwise.

ellyxir:
there's a few uses here, "where binding" and "let binding". the current style seems to favor *where* binding and *let* binding.
there's the keyword itself, that's monospaced and i think thats perfectly fine.
sometimes we say "where clause" or "where block" with different styling. i think "where-block" works the best here without any styling and just use block instead of clause. 
im making a quick patch for this but feel to let me know what you think

ulysses4ever:
Thanks! I'll take a look by the end of the week.

ellyxir:
honestly, i prefer the same thing you do, it just is less consistent with the rest of the book. let me check later today to see what it would take to make that change instead. in this case we'd change it to `where` bindings also, i assume.

MatthijsBlom:
The present diff produces both `where-block` and `*where* bindings`, which feels inconsistent to me.

---

I, too, prefer to always use backticks. Surely mostly personal taste, but maybe it also (ever so slightly) helps me parse: _`where`_ is always code-related, and _where_ always has a common English meaning. It is also easier to get consistent.

For reasons not entirely clear to me, I prefer _`let`-expression_ on the one hand, but _`let` binding_ on the other. If I had to choose, I'd drop the hyphen.

ulysses4ever:
@ellyxir how do you feel about droping the hyphens? I know I proposed to add some initially but now, after @MatthijsBlom brought this up, I think maybe it'll look more consistent without any hyphens, and maybe actually there's no grammatical reason to use them here.

ellyxir:
> @ellyxir how do you feel about droping the hyphens? I know I proposed to add some initially but now, after @MatthijsBlom brought this up, I think maybe it'll look more consistent without any hyphens, and maybe actually there's no grammatical reason to use them here.

works for me, heres the summary of the style as i understand it so far, does it sound right?

### Standalone keyword references
Use monospace (backticks) when referring to the keyword itself. Examples:
- "We put the keyword `where` after the guards"
- "Use a `let` inside a list comprehension"
- "The `let` part is accessible after the `in` part"

### Compound terms with "bindings"
Use monospace for the keyword, followed by a space and "bindings" (no hyphen):
- `where` bindings
- `let` bindings
- `where` binding (singular)
- `let` binding (singular)

### Compound terms with "block" (see note about using with `clause` and `section` also)
Use monospace for the keyword, followed by a space and "block" (no hyphen):
- `where` block
- `where` blocks
- `let` block
Note: "where clause", "where section", and "where block" all refer to the same thing — use `where` block consistently.

ellyxir:
i made the edits, note that there are lots of keywords with italics: do, if, else, case, in, etc... examples:
> That's why we can use *do* syntax to glue together several I/O actions into one.
> Remember that in Haskell, every *if* must have a corresponding *else* because every expression has to have some sort of value.
> Using *case* expressions is commonly used when you want to pattern match against something without bringing in a new function.

i can make those edits too.

ulysses4ever:
@ellyxir 

> works for me, heres the summary of the style as i understand it so far, does it sound right?

yes, thank you for the excellent summary!

> i can make those edits too.

Please, do!

ellyxir:
ok i tried to catch them all

ulysses4ever:
@smith558 @pierluc-codes @MatthijsBlom any more comments here?

MatthijsBlom:
Nope, looks good to me :+1:

source_md/functors-applicative-functors-and-monoids.md
```diff
@@ -1679,7 +1679,7 @@ Another type which can act like a monoid in two distinct but equally valid ways
 The first way is to have the *or* function `||` act as the binary function along with `False` as the identity value.
 The way *or* works in logic is that if any of its two parameters is `True`, it returns `True`, otherwise it returns `False`.
 So if we use `False` as the identity value, it will return `False` when *or*-ed with `False` and `True` when *or*-ed with `True`.
-The `Any` *newtype* constructor is an instance of `Monoid` in this fashion.
+The `Any` `newtype` constructor is an instance of `Monoid` in this fashion.
```
pierluc-codes: If you can reword this sentence to not put two keyword next to each other, that would be amazing

source_md/functors-applicative-functors-and-monoids.md
```diff
@@ -1915,7 +1915,7 @@ instance Monoid (First a) where
 ```
 
 Just like we said.
-`mempty` is just a `Nothing` wrapped with the `First` *newtype* constructor.
+`mempty` is just a `Nothing` wrapped with the `First` `newtype` constructor.
```
pierluc-codes: If you can reword this sentence to not put two keyword next to each other, that would be amazing

pierluc-codes (commented):
I'm very happy with this work. thank you!

source_md/functors-applicative-functors-and-monoids.md
ulysses4ever: Perhaps,

```suggestion
The `newtype` constructor `Any` is an instance of `Monoid` in this fashion.
```

source_md/functors-applicative-functors-and-monoids.md
ulysses4ever:
```suggestion
`mempty` is just a `Nothing` wrapped with the `newtype` constructor `First`.
```

ellyxir:
> Terrific!
> 
> Before merging, it'd be good to cleanup the commit history. The easiest would be just squash into one commit. If you don't have other preferences, I can do it form the github UI and you don't need to bother. A more fine-grained approach would be to have separate commits for the *->' exchange, the hyphen business, and section->block exchange. But it'd probably take more effort than it's worth.

i just made a single commit. it should be ready now.

ellyxir:
feel free to merge ❤️

ulysses4ever:
@ellyxir done, thank you for your perseverance!

PR #134, opened 2026-02-03
Fix minor typo in introduction.md

jet-inigo:
Change "If IRC it not your cup of tea" to "If IRC isn't your cup of tea"

Cheers ☕


ulysses4ever:
the author probably meant "is not". Contractions like "isn't" are considered casual and undesirable in published texts. Although, this text uses many informal constructs. So if people are happy with "isn't", I'd not object. Personally I'd be happier with "is not", @jet-inigo .

jet-inigo:
Just corrected it, thanks @ulysses4ever!

PR #144, opened 2026-03-12
Fix ambiguous type error example

### LINKED ISSUE #25: Example for "ambiguous type variable" error doesn't work anymore
[Chapter 3](https://learnyouahaskell.github.io/types-and-typeclasses.html#typeclasses-101) has an example that `read "4"` produces an "ambiguous type variable" error. However, this is not the case anymore, since GHCi disables the [`MonomorphismRestriction`](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/monomorphism.html#extension-NoMonomorphismRestriction), it now throws a "no parse" exception. The problem is that this is used to explain how to specify a type with an explicit type annotation.

I think the easiest solution would be to simply put the definition in a file. Can you think of a better solution?

MatthijsBlom:
Due to [ExtendedDefaultRules](https://downloads.haskell.org/ghc/latest/docs/users_guide/ghci.html#extension-ExtendedDefaultRules) being enabled by default, GHCi no longer produces an error in the example about ambiguous type variables.

This change amends the original explanation.
It explains the new situation in GHCi, and compares with and elaborates on GHC's behavior.

Fixes #25

- [ ] if b22f71a5277f50edcc7474c60470ddc205bd1026 is accepted: open issue about adding section about type variable defaulting. See https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/144#discussion_r2989733049.

### DIFF
--- a/source_md/types-and-typeclasses.md
+++ b/source_md/types-and-typeclasses.md
@@ -345,18 +345,20 @@ ghci> read "13" :: Bool
 ```
 
 Same as before, and for the same reason.
-The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway -- because we told it to -- and fails.
+The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway --- because we told it to --- and fails.
 
 If you put the same code as before in a file instead of in GHCi, things are a bit different:
 
 ```{.haskell: .ghci}
+-- file: oops.hs
 four = read "4"
 ```
 ```{.haskell: .ghci}
-Oops.hs:4:8: error: [GHC-39999]
+ghci> :l oops.hs
+oops.hs:2:8: error: [GHC-39999]
     • Ambiguous type variable ‘a0’ arising from a use of ‘read’
       prevents the constraint ‘(Read a0)’ from being solved.
-      Relevant bindings include four :: a0 (bound at Oops.hs:4:1)
+      Relevant bindings include four :: a0 (bound at oops.hs:2:1)
       Probable fix: use a type annotation to specify what ‘a0’ should be.
       Potentially matching instances:
         instance Read Ordering -- Defined in ‘GHC.Read’
--- a/source_md/types-and-typeclasses.md
+++ b/source_md/types-and-typeclasses.md
@@ -344,8 +344,9 @@ ghci> read "13" :: Bool
 *** Exception: Prelude.read: no parse
 ```
 
-Same as before, and for the same reason.
-The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway --- because we told it to --- and fails.
+Same as before, and for the same reason: `read` is applied at the wrong type.
+This time not because GHCi guessed the wrong type, but because we explicitly told it to use `Bool`.
+The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway, and fails.
 
 If you put the same code as before in a file instead of in GHCi, things are a bit different:
 
--- a/source_md/types-and-typeclasses.md
+++ b/source_md/types-and-typeclasses.md
@@ -310,6 +310,17 @@ But now, it knows we want some type that is part of the `Read` class, it just do
 And sometimes, when it cannot figure out a type, GHCi tries to be helpful by taking a guess.
 Not just any guess of course, but wrong in this case anyway.
 
+::: {.hintbox}
+**What's this? *Guessing*, at *my* types?!**
+It's not as bad as it sounds!
+But it is a bit of a technical story.
+The process is called *type variable defaulting* and you can read about it in the
+[Type variable defaulting](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_defaulting.html)
+and
+[Type defaulting in GHCi](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/ghci.html#type-defaulting-in-ghci)
+sections of the GHC User's Guide.
+:::
+
 Let's take a look at the type signature of `read`.
 
 ```{.haskell: .ghci}
--- a/source_md/types-and-typeclasses.md
+++ b/source_md/types-and-typeclasses.md
@@ -330,6 +330,12 @@ read :: (Read a) => String -> a
 
 See?
 It returns a type that's part of `Read` but if we don't try to use it in some way later, there's no way of knowing which type.
+Most expressions are such that the compiler can infer what their type is by itself.
+But sometimes, the compiler doesn't know whether to return a value of type `Float` or `[Bool]` for an expression like `read "5"`.
+To see what the type is, Haskell would have to actually evaluate `read "5"`.
+But since Haskell is a statically typed language, it has to know all the types before the code is compiled (or in the case of GHCi, evaluated).
+So we have to tell Haskell: "Hey, this expression should have this type, in case you don't know!".
+
 That's why we can use explicit **type annotations**.
 Type annotations are a way of explicitly saying what the type of an expression should be.
 We do that by adding `::` at the end of the expression and then specifying a type.
@@ -391,12 +397,7 @@ GHC even tells us as much: whatever the type of `four` is, it must be part of `R
 But it can't tell that it is, because it doesn't know the actual type: it is ambiguous.
 Whereas GHCi at this point tried to guess, GHC does not.
 Instead, it immediately complains and demands that we clarify the type.
-
-Most expressions are such that the compiler can infer what their type is by itself.
-But sometimes, the compiler doesn't know whether to return a value of type `Int` or `Float` for an expression like `read "5"`.
-To see what the type is, Haskell would have to actually evaluate `read "5"`.
-But since Haskell is a statically typed language, it has to know all the types before the code is compiled (or in the case of GHCi, evaluated).
-So we have to tell Haskell: "Hey, this expression should have this type, in case you don't know!".
+And this we can do by adding type declarations to definitions, or type annotations to expressions.
 
 `Enum`{.label .class} members are sequentially ordered types --- they can be enumerated.
 The main advantage of the `Enum` typeclass is that we can use its types in list ranges.
--- a/source_md/types-and-typeclasses.md
+++ b/source_md/types-and-typeclasses.md
@@ -297,17 +297,19 @@ But what happens if we try to do just `read "4"`?
 
 ```{.haskell: .ghci}
 ghci> read "4"
-<interactive>:1:0:
-    Ambiguous type variable `a' in the constraint:
-      `Read a' arising from a use of `read' at <interactive>:1:0-7
-    Probable fix: add a type signature that fixes these type variable(s)
+*** Exception: Prelude.read: no parse
 ```
 
-What GHCi is telling us here is that it doesn't know what we want in return.
+Well that's no good.
+What is happening here?
+
 Notice that in the previous uses of `read` we did something with the result afterwards.
 That way, GHCi could infer what kind of result we wanted out of our `read`.
 If we used it as a boolean, it knew it had to return a `Bool`.
 But now, it knows we want some type that is part of the `Read` class, it just doesn't know which one.
+And sometimes, when it cannot figure out a type, GHCi tries to be helpful by taking a guess.
+Not just any guess of course, but wrong in this case anyway.
+
 Let's take a look at the type signature of `read`.
 
 ```{.haskell: .ghci}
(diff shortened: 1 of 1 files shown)


ulysses4ever:
Thank you for submitting this patch! I'll try to review by the end of the week.

ulysses4ever:
In the meantime, it'd be great if you could resolve the conflicts.

source_md/types-and-typeclasses.md
```diff
@@ -335,6 +337,47 @@ ghci> read "(3, 'a')" :: (Int, Char)
...
+4 | four = read "4"
+  |        ^^^^
+```
+
+Yikes!
+Even though this looks very different from earlier, the problem is really the same.
+GHC even tells us as much: whatever the type of `four` is, it must be part of `Read`.
+But it can't tell that it is, because it doesn't know the actual type: it is ambiguous.
+Whereas GHCi at this point tried to guess, GHC does not.
+Instead, it immediately complains and demands that we clarify the type.
```
ulysses4ever, line 399: Can you and an example that type check and works in the end? E.g.

```
> read "4" :: Int
4
```

source_md/types-and-typeclasses.md
```diff
@@ -297,17 +297,19 @@ But what happens if we try to do just `read "4"`?
...
-What GHCI is telling us here is that it doesn't know what we want in return.
+Well that's no good.
+What is happening here?
+
 Notice that in the previous uses of `read` we did something with the result afterwards.
 That way, GHCI could infer what kind of result we wanted out of our `read`.
 If we used it as a boolean, it knew it had to return a `Bool`.
 But now, it knows we want some type that is part of the `Read` class, it just doesn't know which one.
+And sometimes, when it cannot figure out a type, GHCi tries to be helpful by taking a guess.
+Not just any guess of course, but wrong in this case anyway.
```
ulysses4ever, line 311: Have you considered showing what guess it took? It's the unit type. 

```
λ> :set -Wtype-defaults
λ> read "4"
<interactive>:6:1: warning: [GHC-18042] [-Wtype-defaults]
    • Defaulting the type variable ‘a0’ to type ‘()’ in the following constraints
        (Show a0) arising from a use of ‘print’ at <interactive>:6:1-8
        (Read a0) arising from a use of ‘it’ at <interactive>:6:1-8
    • In a stmt of an interactive GHCi command: print it

*** Exception: Prelude.read: no parse
λ> 
```

Maybe mentioning that it's the unit is enough. Otherwise,  ` -Wtype-defaults` may be a good candidate for a NOTE.

source_md/types-and-typeclasses.md
MatthijsBlom: There are already several such examples immediately before this added section.

I tacked on the `read "13" :: Bool` example for contrast with the preceding examples, and also to demonstrate that the _GHCi-guessed-wrong_ explanation is correct.

source_md/types-and-typeclasses.md
MatthijsBlom: I have, but I couldn't find a satisfactory way of figuring it out.
Even already knowing the answer, I failed to trace the exact route that led to it. I thought I remembered a `default` declaration with number types somewhere in `base`, but I couldn't find any `default` declaration that I'm confident is actually used (and remotely relevant).
There seems to be no way of listing all `default` rules in scope.

As it currently stands, nowhere in the book is there any mention of compiler warnings, compiler flags, `:set`, language extensions, or pragmas. Any first mention would likely require quite a bit of extra explanation.

`-Wtype-defaults` seems too specific. But it is included in `-Wall`.
What do you think about writing a suitable `.ghci` and recommending it early in the book?

source_md/types-and-typeclasses.md
```diff
@@ -335,6 +337,47 @@ ghci> read "(3, 'a')" :: (Int, Char)
...
 ```
 
+And what happens if we ask for the wrong type?
+
+```{.haskell: .ghci}
+ghci> read "13" :: Bool
+*** Exception: Prelude.read: no parse
+```
+
+Same as before, and for the same reason.
```
pierluc-codes: That sentence is misleading. Yes it fail with the same error but the reason, under the hood is not the same).

source_md/types-and-typeclasses.md
```diff
@@ -335,6 +337,47 @@ ghci> read "(3, 'a')" :: (Int, Char)
...
+The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway -- because we told it to -- and fails.
+
+If you put the same code as before in a file instead of in GHCi, things are a bit different:
+
+```{.haskell: .ghci}
+four = read "4"
+```
+```{.haskell: .ghci}
+Oops.hs:4:8: error: [GHC-39999]
```
pierluc-codes: the line number is wrong. worth to clarify.

also, i think it would be great if you can put the same of the file as a comment in the previous snippet (that is my favorite technique, i am open to alternative). It will make the example easier to parse.

source_md/types-and-typeclasses.md
pierluc-codes: > And sometimes, when it cannot figure out a type, GHCi tries to be helpful by taking a guess.
> Not just any guess of course, but wrong in this case anyway.

IMO, this is a little bit too vague, because it makes the reader feel that ghci behavior is a surprise. It is kind of tricky to clarify because () was not introduced yet. I am not totally sure how to solve this tension.

source_md/types-and-typeclasses.md
```diff
@@ -335,6 +337,47 @@ ghci> read "(3, 'a')" :: (Int, Char)
...
+
+Same as before, and for the same reason.
+The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway -- because we told it to -- and fails.
+
+If you put the same code as before in a file instead of in GHCi, things are a bit different:
+
+```{.haskell: .ghci}
+four = read "4"
+```
+```{.haskell: .ghci}
```
pierluc-codes, line 374: This is more a shell output dans ghci.

source_md/types-and-typeclasses.md
```diff
@@ -335,6 +337,47 @@ ghci> read "(3, 'a')" :: (Int, Char)
...
+```
+
+Yikes!
+Even though this looks very different from earlier, the problem is really the same.
+GHC even tells us as much: whatever the type of `four` is, it must be part of `Read`.
+But it can't tell that it is, because it doesn't know the actual type: it is ambiguous.
+Whereas GHCi at this point tried to guess, GHC does not.
+Instead, it immediately complains and demands that we clarify the type.
+
 Most expressions are such that the compiler can infer what their type is by itself.
```
pierluc-codes: I _think_ this part should be moved.

source_md/types-and-typeclasses.md
```diff
@@ -335,6 +337,47 @@ ghci> read "(3, 'a')" :: (Int, Char)
...
 
+And what happens if we ask for the wrong type?
+
+```{.haskell: .ghci}
+ghci> read "13" :: Bool
+*** Exception: Prelude.read: no parse
+```
+
+Same as before, and for the same reason.
+The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway -- because we told it to -- and fails.
```
pierluc-codes: I think the rest of the document uses `---`

pierluc-codes (commented):
My apologies, I put a lot of comment on your work: do not take that as a criticism of the quality of the work. I really appreciate your changes here. 

I especially appreciate the effort of keeping the tone of the rest of the book.

Thank you!

source_md/types-and-typeclasses.md
MatthijsBlom: The common reason I had in mind was `read` being applied at the wrong type. Are you instead thinking of there (not) having been a guess, or of something else?

source_md/types-and-typeclasses.md
MatthijsBlom: It does. I'll change these to match.

source_md/types-and-typeclasses.md
MatthijsBlom: Elsewhere in the book the reader is instructed to `:load` their files into GHCi.

I didn't think to previously, but I think I'll add a `ghci> :l oops.hs` line, to be more explicit.

There are three other error messages in the book. You can find them by searching for `•`. One is labeled `{.haskell:hs}`, the other two `{.haskell: .ghci}`. Unaware of the other one, I used the latter because of how it renders. Is there a more suitable option available?

source_md/types-and-typeclasses.md
MatthijsBlom: Ideally, the snippet as a whole could be labeled.

I have added a filename comment for now. It'll place the error on different lines for different readers, depending on whether they copy the comment or not. But I don't think this is important.

source_md/types-and-typeclasses.md
MatthijsBlom: Yeah, I considered doing this as well. In the end I kept it because of its final line, because I thought the repetition wouldn't hurt, and because it seemed to flow better into the next section.

I'll do some thinking about how to incorporate the last line's information somewhere else.

source_md/types-and-typeclasses.md
MatthijsBlom: But it _is_ a surprise, isn't it? :smirk:
I'd prefer GHC(i) just didn't do this by default.

`()` is in fact introduced earlier in this same chapter.
I don't think the bulk of the trickiness lies with `()`. But it sure is tricky.
I'll do some more thinking.

[errors.haskell.org/messages/GHC-39999](https://errors.haskell.org/messages/GHC-39999/) seems outdated, by the way.

MatthijsBlom:
> My apologies, I put a lot of comment on your work: do not take that as a criticism of the quality of the work. I really appreciate your changes here.

No worries!
 
> I especially appreciate the effort of keeping the tone of the rest of the book.

Good to hear that this has apparently been successful.

source_md/types-and-typeclasses.md
pierluc-codes: in read "4" the type is ambiguous, the guess is wrong.
in read "13" :: Bool, the type is unambigious but unparsable.

Not exactly the same reason :)

source_md/types-and-typeclasses.md
pierluc-codes: Thank you.

source_md/types-and-typeclasses.md
pierluc-codes: Just for my curosity: how is the render with bash/shellscript? (Only curious, I am not suggesting that as a fix)

source_md/types-and-typeclasses.md
MatthijsBlom: I think my preference would be to add a `.hintbox` here with a reference to a yet-to-be-written appendix specifically about type variable defaulting.

There is a lot to be explained (extensions/settings, GHCi vs. GHC, default rules, the defaulting mechanisms, possibly warnings, possibly interaction with the monomorphism restriction), and probably several examples should be included. Including all this in this chapter would, I feel, disrupt the text too much. But defaulting is still worth discussing. Therefore: split it off.

What do you think?

If this way is chosen, I'd prefer to defer adding this appendix to a future PR.

source_md/types-and-typeclasses.md
MatthijsBlom: What are your thoughts on
```markdown
Same as before, and for the same reason: `read` is applied at the wrong type.
This time not because GHCi guessed the wrong type, but because we explicitly told it to use `Bool`.
The string `"13"` cannot be parsed as a `Bool`, but `read` tries to anyway, and fails.
```

source_md/types-and-typeclasses.md
MatthijsBlom: This is either of `{.haskell: .ghci}` and `{.haskell:hs}`:

Image

This is `{.plain}`:

Image

source_md/types-and-typeclasses.md
pierluc-codes: Sure, perfect is the enemy of good.

source_md/types-and-typeclasses.md
pierluc-codes: Yep, look great :+1:

source_md/types-and-typeclasses.md
pierluc-codes: Thank you!

source_md/types-and-typeclasses.md
MatthijsBlom: I added [b22f71a](https://github.com/learnyouahaskell/learnyouahaskell.github.io/pull/144/commits/b22f71a5277f50edcc7474c60470ddc205bd1026) for now.

source_md/types-and-typeclasses.md
MatthijsBlom: Addressed in 73a8fd1914d99678fd6adf57921a5691c7e63d43.

source_md/types-and-typeclasses.md
MatthijsBlom: Do you still feel there should be an example of success further down?

ulysses4ever:
I promise I'll get to this one this month of April! Sorry for the delay!

PR #146, opened 2026-03-15
`Monoid` update

### LINKED ISSUE #49: "Monoids" is outdated
In: [Functors, Applicative Functors and Monoids](https://learnyouahaskell.github.io/functors-applicative-functors-and-monoids.html) it described Monoids as having a minimal definition of `mempty` and `mappend`. Nowadays one has to define the Semigroup first and then define just `mempty` to create a Monoid. So people should not manually define `mappend` anymore. Perhaps a section on Semigroups can be created out of the existing sectiom for Monoids?

MatthijsBlom:
Fix #49

I recommend reviewing the commits individually, in order.
As some changes' contexts are bigger than GitHub's UI suggests, consider reviewing locally.

### DIFF
--- a/source_md/functors-applicative-functors-and-monoids.md
+++ b/source_md/functors-applicative-functors-and-monoids.md
@@ -1486,18 +1486,20 @@ We call this property *associativity*.
 `*` is associative, and so is `++`, but `-`, for example, is not.
 The expressions `(5 - 3) - 4` and `5 - (3 - 4)` result in different numbers.
 
-By noticing and writing down these properties, we have chanced upon *monoids*!
-A monoid is when you have an associative binary function and a value which acts as an identity with respect to that function.
+By noticing and writing down these properties, we have chanced upon *semigroups* and *monoids*!
+A semigroup is when you have an associative binary function, and a monoid is when you also have a value which acts as an identity with respect to that function.
 When something acts as an identity with respect to a function, it means that when called with that function and some other value, the result is always equal to that other value.
 `1` is the identity with respect to `*` and `[]` is the identity with respect to `++`.
-There are a lot of other monoids to be found in the world of Haskell, which is why the `Monoid` type class exists.
-It's for types which can act like monoids.
-Let's see how the type class is defined:
+There are a lot of other monoids and semigroups to be found in the world of Haskell, which is why the `Monoid` and `Semigroup` type classes exist.
+It's for types which can act like monoids and semigroups.
+Let's see how the type classes are defined:
 
 ```{.haskell:hs}
-class Monoid m where
-    mempty :: m
+class Semigroup m where
     (<>) :: m -> m -> m
+
+class Semigroup m => Monoid m where
+    mempty :: m
     mconcat :: [m] -> m
     mconcat = foldr (<>) mempty
 ```
@@ -1508,6 +1510,8 @@ Let's take some time and get properly acquainted with it.
 
 First of all, we see that only concrete types can be made instances of `Monoid`, because the `m` in the type class definition doesn't take any type parameters.
 This is different from `Functor` and `Applicative`, which require their instances to be type constructors which take one parameter.
+We also see that all `Monoid`s must also be `Semigroup`s, just like all `Applicative`s must be `Functor`s.
+This is because monoids are a special kind of semigroup, as they also have an identity value.
 
 The first function is `mempty`.
 It's not really a function, since it doesn't take parameters, so it's a polymorphic constant, kind of like `minBound` from `Bounded`.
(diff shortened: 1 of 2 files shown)


ulysses4ever (approved):
Thanks! And sorry for a long wait.

I think it's mergeable although I'd make different choices sometimes.

First, reviewing commit-by-commit on GitHub is hard. While it is true that reading your changes commit-by-commit makes more sense (and thank you for suggesting it),   commenting those changes is another story. Comments you leave under a certain commit (as opposed to in the overall "Files changed" tab of a PR) are second-class citizens in certain ways making them really hard to track. So, I'd much prefer you did the mappend->`<>` substitution in a separate PR (I know it may make little sense on its own).

Second, my main pain point here is in textual changes: sometimes it feels like the new text spells out "monoid and semigroup" where just "monoid" would suffice, and this makes it harder to read. For instance,

> It's possible to make instances of `Semigroup` and `Monoid` that don't follow these rules,

I think it could do with just "instances of `Monoid`". Especially given that just above it does say talk only about Monoids ("Before moving on to specific instances of `Monoid`"), and that's the style I 'd prefer because every time use say something about monoids this usually implies things about semigroups and unless there's a vbery particular need for mentioning semigoroups, I'd prefer to just read about monoids as a whole. But I understand that it's a stylistic choice, so I don't want to argue about it...

On the other hand, where old text talks about code specifically, I feel, it should mention both instances instead of one. E.g.

> Here's the `Monoid` instance:

```{.haskell:hs}
instance Semigroup (DiffList a) where
    DiffList f <> DiffList g = DiffList (\xs -> f (g xs))
instance Monoid (DiffList a) where
    mempty = DiffList (\xs -> [] ++ xs)
    (DiffList f) `mappend` (DiffList g) = DiffList (\xs -> f (g xs))
```

This doesn't make sense: it says "here's one instance" and shows two. It must say that two instancces are coming.

MatthijsBlom:
Good to know reviewing individual commits on here is that annoying.

> So, I'd much prefer you did the mappend->`<>` substitution in a separate PR (I know it may make little sense on its own).

I have extracted the first two commits into #154.

---

> But I understand that it's a stylistic choice, so I don't want to argue about it...

The main reason I chose to mention `Semigroup` so often was to provide the reader with extra exposure to this new concept. They have likely never encountered it before, and the semigroup concept is easily overshadowed by that of the monoid.
I agree it makes the text harder to read. It was a tradeoff. I'll go over the text again after posting this reply to see how I can improve the text's flow.

source_md/functors-applicative-functors-and-monoids.md
```diff
@@ -1760,16 +1773,18 @@ With lists, numbers and boolean values, finding monoids was just a matter of loo
 With `Ordering`, we have to look a bit harder to recognize a monoid, but it turns out that its `Monoid` instance is just as intuitive as the ones we've met so far and also quite useful:
```
MatthijsBlom: This seems fine to me, but could also be
```suggestion
With `Ordering`, we have to look a bit harder to recognize a monoid, but it turns out that its  instances are just as intuitive as the ones we've met so far and also quite useful:
```

MatthijsBlom (commented):
I rebased onto #154. Hopefully that will improve the diff when that PR is merged.

Feel free to nitpick.
After everything is settled I plan to squash into sensible chunks.

source_md/functors-applicative-functors-and-monoids.md
ulysses4ever: I'd prefer the current version (explicitly mentioning just Monoid).

source_md/functors-applicative-functors-and-monoids.md
```diff
@@ -1486,18 +1486,20 @@ We call this property *associativity*.
 `*` is associative, and so is `++`, but `-`, for example, is not.
 The expressions `(5 - 3) - 4` and `5 - (3 - 4)` result in different numbers.
 
-By noticing and writing down these properties, we have chanced upon *monoids*!
-A monoid is when you have an associative binary function and a value which acts as an identity with respect to that function.
+By noticing and writing down these properties, we have chanced upon *semigroups* and *monoids*!
+A semigroup is when you have an associative binary function, and a monoid is when you also have a value which acts as an identity with respect to that function.
```
ulysses4ever, line 1490:
```suggestion
A semigroup is when you have an associative binary function, and a monoid is when you have a semigroup and a value that acts as an identity with respect to that function.
```

source_md/functors-applicative-functors-and-monoids.md
```diff
@@ -1486,18 +1486,20 @@ We call this property *associativity*.
...
 When something acts as an identity with respect to a function, it means that when called with that function and some other value, the result is always equal to that other value.
 `1` is the identity with respect to `*` and `[]` is the identity with respect to `++`.
-There are a lot of other monoids to be found in the world of Haskell, which is why the `Monoid` type class exists.
-It's for types which can act like monoids.
-Let's see how the type class is defined:
+There are a lot of other monoids and semigroups to be found in the world of Haskell, which is why the `Monoid` and `Semigroup` type classes exist.
```
ulysses4ever, line 1493:
```suggestion
There are a lot of other monoids and semigroups to be found in the world of Haskell, which is why the respective type classes (`Semigroup` and `Monoid`) exist.
```
would read a little cleaner I feel.

source_md/functors-applicative-functors-and-monoids.md
```diff
@@ -1486,18 +1486,20 @@ We call this property *associativity*.
...
+It's for types which can act like monoids and semigroups.
```
ulysses4ever, line 1494: The ordering in "monoids and semigroups" is a bit confusing, I feel, because semigroups come first in the sense that any monoid is a semigroup as well. But maybe it's not a big deal. Just in case, I add a suggestion but it's up to you whether to acccept it.
```suggestion
It's for types which can act like semigroups and monoids.
```

ulysses4ever (approved):
Thanks! A tiny bit of ideas inline but it's already quite good!

source_md/functors-applicative-functors-and-monoids.md
ulysses4ever: Having just "also" there doesn't pull the weight I feel. It's important to be crystal clear about this hierarchy.

MatthijsBlom: The "that function" reference feels broken, that way.

Maybe a total rewrite?

```suggestion
When you have an associative binary function, we call that a semigroup.
And if you also have a value that acts as an identity with respect to that function, then we call it a monoid instead.
```

( The 'instead' is there to make clear that it is the whole that is called the 'monoid', not just the identity element by itself. )

source_md/functors-applicative-functors-and-monoids.md
MatthijsBlom: I would agree, if this weren't the first time these classes were named/mentioned.

source_md/functors-applicative-functors-and-monoids.md
MatthijsBlom: I do not expect the reader's intuition to be clear on this yet, at this point in the text, so I do not expect a problem.

If these two words are to be swapped, I think the preceding two occurrences should be swapped as well, to preserve the 'rhythm'.

Currently the 'most important' of the two comes first, in these two sentences, somewhat in agreement with the title of this section.

PR #148, opened 2026-03-24
Remove outdated Error type class and Either Monad constraints

Xingye-Dujing:
This PR removes outdated content regarding the `Error` type class and the constrained `Monad (Either e)` instance.

In modern Haskell:
- `Either e` is an unconditional instance of `Monad` (no `Error e` constraint required).
- Based on the old `fail` method residing in the `Monad` type class, which has since been moved to `MonadFail`. Pattern match failures in `do` blocks involving `Either` no longer automatically convert strings to `Left` values via `strMsg`; instead, they require a `MonadFail` instance or explicit handling.
- `Right 3 >>= \x -> return (x + 100)` no longer throws an error.

image

## References
- [GHC.Internal.Data.Either](https://hackage-content.haskell.org/package/ghc-internal-9.1401.0/docs/GHC-Internal-Data-Either.html#v:-62--62--61-)
- [Control.Monad.Fail](https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Monad-Fail.html)

### DIFF
--- a/source_md/for-a-few-monads-more.md
+++ b/source_md/for-a-few-monads-more.md
@@ -1087,11 +1087,10 @@ This is pretty much just an enhanced `Maybe`, so it makes sense for it to be a m
 Its `Monad` instance is similar to that of `Maybe` and it can be found in `Control.Monad.Error`:
 
 ```{.haskell:hs}
-instance (Error e) => Monad (Either e) where
+instance Monad (Either e) where
     return x = Right x
     Right x >>= f = f x
     Left err >>= f = Left err
-    fail msg = Left (strMsg msg)
 ```
 
 `return`, as always, takes a value and puts it in a default minimal context.
@@ -1102,20 +1101,6 @@ The `>>=` examines two possible cases: a `Left` and a `Right`.
 In the case of a `Right`, the function `f` is applied to the value inside it, similar to how in the case of a `Just`, the function is just applied to its contents.
 In the case of an error, the `Left` value is kept, along with its contents, which describe the failure.
 
-The `Monad` instance for `Either e` makes an additional requirement, and that is that the type of the value contained in a `Left`, the one that's indexed by the `e` type parameter, has to be an instance of the `Error` type class.
-The `Error` type class is for types whose values can act like error messages.
-It defines the `strMsg` function, which takes an error in the form of a string and returns such a value.
-A good example of an `Error` instance is, well, the `String` type!
-In the case of `String`, the `strMsg` function just returns the string that it got:
-
-```{.haskell:hs}
-ghci> :t strMsg
-strMsg :: (Error a) => String -> a
-ghci> strMsg "boom!" :: String
-"boom!"
-```
-
-But since we usually use `String` to describe the error when using `Either`, we don't have to worry about this too much.
 When a pattern match fails in `do` notation, a `Left` value is used to signify this failure.
 
 Anyway, here are a few examples of usage:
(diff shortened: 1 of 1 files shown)


ulysses4ever (approved):
Thanks! Let's wait if someone else has an opinion but it looks good to me

source_md/for-a-few-monads-more.md
```diff
@@ -1087,11 +1087,10 @@ This is pretty much just an enhanced `Maybe`, so it makes sense for it to be a m
 Its `Monad` instance is similar to that of `Maybe` and it can be found in `Control.Monad.Error`:
```
MatthijsBlom: `Control.Monad.Error` does not exist anymore.

source_md/for-a-few-monads-more.md
```diff
@@ -1102,20 +1101,6 @@ The `>>=` examines two possible cases: a `Left` and a `Right`.
...
-
-```{.haskell:hs}
-ghci> :t strMsg
-strMsg :: (Error a) => String -> a
-ghci> strMsg "boom!" :: String
-"boom!"
-```
-
-But since we usually use `String` to describe the error when using `Either`, we don't have to worry about this too much.
 When a pattern match fails in `do` notation, a `Left` value is used to signify this failure.
```
MatthijsBlom: This does not happen, as `Either _` is not `MonadFail`.

```haskell
_ = do
  False <- pure True
  Right ()
```

source_md/for-a-few-monads-more.md
```diff
@@ -1130,33 +1115,9 @@ Left "no way!"
...
-Right 103
-```
-
-Alright, now it works!
-
-Other than this little hangup, using this monad is very similar to using `Maybe` as a monad.
+Using this monad is very similar to using `Maybe` as a monad.
 In the previous chapter, we used the monadic aspects of `Maybe` to simulate birds landing on the balancing pole of a tightrope walker.
-As an exercise, you can rewrite that with the error monad so that when the tightrope walker slips and falls, we remember how many birds were on each side of the pole when he fell.
+As an exercise, you can rewrite that with the `Either` monad so that when the tightrope walker slips and falls, we remember how many birds were on each side of the pole when he fell.
```
MatthijsBlom:
```suggestion
As an exercise, you can rewrite that with `Either` so that when the tightrope walker slips and falls, we remember how many birds were on each side of the pole when he fell.
```
Phrases like "the _ monad" are unhealthy. See [Stop saying "the ___ monad"](https://www.reddit.com/r/haskell/comments/24kvwh/evan_czaplicki_stop_saying_the_monad/), [IO Monad Considered Harmful](https://blog.jle.im/entry/io-monad-considered-harmful.html)

source_md/for-a-few-monads-more.md
Xingye-Dujing: Yes, I will change `it can be found in Control.Monad.Error` to `it can be found in Data.Either`.

[Data.Either](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Either.html)

source_md/for-a-few-monads-more.md
Xingye-Dujing: Thanks for catching that! Would it be better if I updated the text to say: `Note that unlike Maybe, pattern match failures in do notation will result in a runtime exception, not a Left value.`?

source_md/for-a-few-monads-more.md
Xingye-Dujing: Got it. I'll go with your revised version.

source_md/for-a-few-monads-more.md
MatthijsBlom: The instance is provided by the `Prelude`. Perhaps this was different back then? In any case, today both `Prelude` and `Data.Either` reexport it from `GHC.Internal.Data.Either`.

None of `Data.Either`'s non-`Prelude` exports are discussed in the book. I think it can be left out. `Data.Maybe` isn't mentioned anywhere either.

Xingye-Dujing: I will follow your instructions.

source_md/for-a-few-monads-more.md
```diff
@@ -1102,21 +1101,7 @@ The `>>=` examines two possible cases: a `Left` and a `Right`.
...
-```{.haskell:hs}
-ghci> :t strMsg
-strMsg :: (Error a) => String -> a
-ghci> strMsg "boom!" :: String
-"boom!"
-```
-
-But since we usually use `String` to describe the error when using `Either`, we don't have to worry about this too much.
-When a pattern match fails in `do` notation, a `Left` value is used to signify this failure.
+Note that unlike Maybe, pattern match failures in do notation will result in a runtime exception, not a Left value.
```
MatthijsBlom:
```suggestion
Note that unlike `Maybe`, pattern match failures in `do` notation will result in a runtime exception, not a `Left` value.
```

Thoughts, but not necessarily suggestions:

- Exceptions aren't really discussed anywhere in the book.
- This remark could use an accompanying example,
- but this could also be revisited when the book is updated on `MonadFail`,
- which I've been thinking about maybe working on sometime soon.

Xingye-Dujing: Let's hold off on adding the example until the book is updated to cover `MonadFail`. Currently, the book doesn't mention `MonadFail` at all, so including an example now would make its appearance feel somewhat abrupt.

ulysses4ever:
@MatthijsBlom any last thoughts? Is it good to go?

MatthijsBlom:
Looks all good to me :+1:

ulysses4ever:
Hmm, somehow, this one did merge get in even though it had a merge commit. Confusing…

PR #155, opened 2026-04-08
Use MathML for math formatting

MatthijsBlom:
For now this is only a proof of concept.
If this direction is appreciated, I might look into it further.

To Do:

- [ ] test on a variety of clients
- [ ] find a more suitable math font

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |
|  |  |


pierluc-codes:
I like it! 

For few second, I was hesitant because the previous syntax was closer to code. But, we are talking math, so the visual make sense.

If you can upload it somewhere, I can test on my iDevice and check if render similar.

Thanks!

source_md/recursion.md
```diff
@@ -17,14 +17,14 @@ Recursion is actually a way of defining functions in which the function is appli
 Definitions in mathematics are often given recursively.
 For instance, the fibonacci sequence is defined recursively.
 First, we define the first two fibonacci numbers non-recursively.
-We say that *F(0) = 0* and *F(1) = 1*, meaning that the 0th and 1st fibonacci numbers are 0 and 1, respectively.
+We say that $F(0) = 0$ and $F(1) = 1$, meaning that the 0<sup>th</sup> and 1<sup>st</sup> fibonacci numbers are $0$ and $1$, respectively.
```
ulysses4ever: there's a Pandoc option for superscripts: https://pandoc.org/MANUAL.html#extension-superscript-subscript

source_md/starting-out.md
```diff
@@ -884,7 +884,7 @@ ghci> triangles = [ (a,b,c) | c <- [1..10], a <- [1..10], b <- [1..10] ]
 We're just drawing from three lists and our output function is combining them into a triple.
 If you evaluate that by typing out `triangles` in GHCi, you'll get a list of all possible triangles with sides under or equal to 10.
 Next, we'll add a condition that they all have to be right triangles.
-We'll also modify this function by taking into consideration that side *a* isn't larger than the hypotenuse and that side *b* isn't larger than side *a*.
+We'll also modify this function by taking into consideration that side $a$ isn't larger than the hypotenuse and that side $b$ isn't larger than side $a$.
```
ulysses4ever, line 887: is there a good reason to switch from * to $? the former looks simpler to me and I naturally prefer simpler approaches.

source_md/higher-order-functions.md
```diff
@@ -865,7 +865,7 @@ ghci> map ($ 3) [(4+), (10*), (^2), sqrt]
 
 ## Function composition {#composition}
 
-In mathematics, function composition is defined like this: ![ (f . g)(x) = f(g(x))](assets/images/higher-order-functions/composition.png), meaning that composing two functions produces a new function that, when called with a parameter, say, *x* is the equivalent of calling *g* with the parameter *x* and then calling the *f* with that result.
+In mathematics, function composition is defined like this: $\left( f \circ g \right) (x) = f \left( g(x) \right)$, meaning that composing two functions produces a new function that, when called with an argument, say, $x$ is the equivalent of calling $g$ with the argument $x$ and then calling the $f$ with that result.
```
ulysses4ever, line 868: do you really need to use \left and \right in these?

ulysses4ever (approved):
Great contribution, thanks!

Small comments inline.

source_md/higher-order-functions.md
MatthijsBlom: No, but it looks slightly nicer:
image
Though still not as nice as I'd like. Ideally, the parentheses would share a horizontal symmetry access, and the bigger ones would therefore cross the baseline.

source_md/starting-out.md
MatthijsBlom: No, this one is really up to personal preferences.

source_md/recursion.md
MatthijsBlom: Ah, so that's why there were these paired `^`s in the polynomials.

source_md/higher-order-functions.md
ulysses4ever: I personally prefer same-sized parenthesis for expressions as simple as these for uniformity sake. I only resort to \left and \right when the expressions are big and it's hard to match the parenthesis. But I don't feel too strong about it in this case.

MatthijsBlom:
> If you can upload it somewhere, I can test on my iDevice and check if render similar.

This branch is now live at [matthijsblom.github.io/learnyouahaskell.github.io](https://matthijsblom.github.io/learnyouahaskell.github.io/).

Need I fear that the changed build script (to be reverted before merge) on this branch will cause spurious builds in this repo? We don't want this branch served at learnyouahaskell.github.io.

ulysses4ever:
Off-topic: We really should implement automatic PR preview. But last time I looked at it recently, is a bit of a pain in the neck to set up.

pierluc-codes:
> This branch is now live at [matthijsblom.github.io/learnyouahaskell.github.io](https://matthijsblom.github.io/learnyouahaskell.github.io/).

All good on my iDevice. Thank you!

pierluc-codes:
@MatthijsBlom 

If you can revert the github action change, remove the draft then I will approve. Thank you!

source_md/higher-order-functions.md
ulysses4ever: all right, since it's not a big deal, let's just agree to disagree :-) resolving...

ulysses4ever (commented):
LGTM modulo the workflow edits to deploy locally

MatthijsBlom:
I _think_ I've covered the whole book now.

I don't think the `$…$`-ification of single names is a big improvement. But using `` `…` `` or `*…*` isn't great either, nor is not styling at all. My preference for `$…$` is very mild. In any case, let's be consistent.

I'll rebase to improve the commit messages and remove the CI change after incorporating the final feedback.

( By the way: I've been doing a pass through the whole book. If you're curious, see [my `dump` branch](https://github.com/learnyouahaskell/learnyouahaskell.github.io/compare/main...MatthijsBlom:learnyouahaskell.github.io:dump). Not everything on there is meant for eventual merging. )

ulysses4ever:
I'm fine with $'s for single identifiers. I resolved all my comments from earlier, so to my mind this is ready to go.

PR #157, opened 2026-04-26
Kinds update

MatthijsBlom:
GHC 9.6.7 (currently recommended by GHCup) uses `GHC2021` as the default language, which includes `PolyKinds`.
This renders the `:kind` output shown in the book out of date.

I include an additional section on `:k Functor` because I feel it helps ground the rest of the discussion. With it, `Functor`'s requirements are no longer only to be inferred, but also plain to see in the output of `:k`.

The `[Char]` &mapsto; `String` change is included because GHCi gives the latter (because string literals are used in the input).

### DIFF
--- a/source_md/making-our-own-types-and-typeclasses.md
+++ b/source_md/making-our-own-types-and-typeclasses.md
@@ -1811,6 +1811,22 @@ ghci> :k Either String Int
 Either String Int :: *
 ```
 
+What if we do something crazy and ask what the kind is of something entirely different, like a type class?
+
+```{.haskell:hs}
+ghci> :k Functor
+Functor :: (* -> *) -> Constraint
+```
+
+Woot, that works!
+We see two new things here.
+First, `Constraint` is the kind of constraints, i.e. the things that appear to the left of the `=>` in type signatures.
+The `Ord Int` instance, for instance, has kind `Constraint`.
+
+We also see that, whereas `Maybe` and `Either` take concrete types as arguments, `Functor` does not.
+Instead of a concrete type, it takes something that *itself* takes a single concrete type as an argument.
+So it can take `Maybe`, which has kind `* -> *`, but it will not take `Char`, as its kind is merely `*`.
+
 When we wanted to make `Either` a part of the `Functor` typeclass, we had to partially apply it because `Functor` wants types that take only one parameter while `Either` takes two.
 In other words, `Functor` wants types of kind `* -> *` and so we had to partially apply `Either` to get a type of kind `* -> *` instead of its original kind `* -> * -> *`.
 If we look at the definition of `Functor` again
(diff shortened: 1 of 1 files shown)


source_md/making-our-own-types-and-typeclasses.md
```diff
@@ -1811,6 +1811,22 @@ ghci> :k Either String Int
...
+
+```{.haskell:hs}
+ghci> :k Functor
+Functor :: (* -> *) -> Constraint
+```
+
+Woot, that works!
+We see two new things here.
+First, `Constraint` is the kind of constraints, i.e. the things that appear to the left of the `=>` in type signatures.
+The `Ord Int` instance, for instance, has kind `Constraint`.
```
pierluc-codes: This is a little bit imprecise: the instance is the implementation that satisfies the constaints. It's `Ord Int` itself that has kind Constraint, not the instance.

source_md/making-our-own-types-and-typeclasses.md
```diff
@@ -1892,32 +1911,32 @@ Let's do some more type-foo.
...
-Barry :: (* -> *) -> * -> * -> *
+Barry :: (k -> *) -> k -> * -> *
 ```
 
 Ah, we were right.
 How satisfying.
-Now, to make this type a part of `Functor` we have to partially apply the first two type parameters so that we're left with `* -> *`.
-That means that the start of the instance declaration will be: `instance Functor (Barry a b) where`.
+Now, to make this type a part of `Functor` we have to fill the first two parameters so that we're left with `* -> *`.
+That means that the start of the instance declaration will be: `instance Functor (Barry … …) where`.
```
pierluc-codes, line 1934: Any reason for the ellipsis?

pierluc-codes (commented):
small detail to fix. thank you!

source_md/making-our-own-types-and-typeclasses.md
MatthijsBlom: Of course 🤦

I think I confused myself by thinking of the implicitly passed dictionary, and by considering constraints singleton 'types'.

source_md/making-our-own-types-and-typeclasses.md
MatthijsBlom: The same names `a` and `b` are used for something else in the very next sentence. Justification for using `c` and `d` instead would come only after this sentence, which felt a bit weird, and without justification it would be arbitrary. Using <code>_something_</code> felt too heavy / noisy. I considered using `_`, but that being valid syntax might lead to some confusion. In the end I figured I'd try and see how reviewers would feel about `…`. It's the shape of the instance type that is the point; arbitrary variable names need no drawing attention to.

MatthijsBlom:
( Lemme squash the feedback commits before merging please. )

PR #158, opened 2026-05-02
Minor general fixes

talestonini:
Mostly English fixes and a small Haskell syntax fix.

### DIFF
--- a/source_md/higher-order-functions.md
+++ b/source_md/higher-order-functions.md
@@ -865,7 +865,7 @@ ghci> map ($ 3) [(4+), (10*), (^2), sqrt]
 
 ## Function composition {#composition}
 
-In mathematics, function composition is defined like this: ![ (f . g)(x) = f(g(x))](assets/images/higher-order-functions/composition.png), meaning that composing two functions produces a new function that, when called with a parameter, say *x*, is the equivalent of calling *g* with the parameter *x* and then calling the *f* with that result.
+In mathematics, function composition is defined like this: ![ (f . g)(x) = f(g(x))](assets/images/higher-order-functions/composition.png), meaning that composing two functions produces a new function that, when called with a parameter, let's say *x*, is the equivalent of calling *g* with the parameter *x* and then calling the *f* with that result.
 
 In Haskell, function composition is pretty much the same thing.
 We do function composition with the `.` function, which is defined like so:
--- a/source_md/higher-order-functions.md
+++ b/source_md/higher-order-functions.md
@@ -340,7 +340,7 @@ There's no set rule for when to use `map` and `filter` versus using list compreh
 The `filter` equivalent of applying several predicates in a list comprehension is either filtering something several times or joining the predicates with the logical `&&` function.
 
 Remember our quicksort function from the [previous chapter](recursion.html)?
-We used list comprehensions to filter out the list of elements that are smaller than (or equal to) and larger than the pivot.
+We used list comprehensions to filter out the list elements that are smaller than (or equal to) and larger than the pivot.
 We can achieve the same functionality in a more readable way by using `filter`:
 
 ```{.haskell:ghci}
--- a/source_md/higher-order-functions.md
+++ b/source_md/higher-order-functions.md
@@ -207,7 +207,7 @@ zipWith' f (x:xs) (y:ys) = f x y : zipWith' f xs ys
 Look at the type declaration.
 The first parameter is a function that takes two things and produces a third thing.
 They don't have to be of the same type, but they can.
-The second and third parameter are lists.
+The second and third parameters are lists.
 The result is also a list.
 The first has to be a list of `a`'s, because the joining function takes `a`'s as its first argument.
 The second has to be a list of `b`'s, because the second parameter of the joining function is of type `b`.
@@ -340,7 +340,7 @@ There's no set rule for when to use `map` and `filter` versus using list compreh
 The `filter` equivalent of applying several predicates in a list comprehension is either filtering something several times or joining the predicates with the logical `&&` function.
 
 Remember our quicksort function from the [previous chapter](recursion.html)?
-We used list comprehensions to filter out the list elements that are smaller than (or equal to) and larger than the pivot.
+We used list comprehensions to filter out the list of elements that are smaller than (or equal to) and larger than the pivot.
 We can achieve the same functionality in a more readable way by using `filter`:
 
 ```{.haskell:ghci}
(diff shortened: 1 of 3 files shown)


source_md/higher-order-functions.md
```diff
@@ -340,7 +340,7 @@ There's no set rule for when to use `map` and `filter` versus using list compreh
 The `filter` equivalent of applying several predicates in a list comprehension is either filtering something several times or joining the predicates with the logical `&&` function.
 
 Remember our quicksort function from the [previous chapter](recursion.html)?
-We used list comprehensions to filter out the list elements that are smaller than (or equal to) and larger than the pivot.
+We used list comprehensions to filter out the list of elements that are smaller than (or equal to) and larger than the pivot.
```
ulysses4ever: Here, it actually means "list elements" rather than "list of elements" I think.

source_md/higher-order-functions.md
```diff
@@ -865,7 +865,7 @@ ghci> map ($ 3) [(4+), (10*), (^2), sqrt]
 
 ## Function composition {#composition}
 
-In mathematics, function composition is defined like this: ![ (f . g)(x) = f(g(x))](assets/images/higher-order-functions/composition.png), meaning that composing two functions produces a new function that, when called with a parameter, say, *x* is the equivalent of calling *g* with the parameter *x* and then calling the *f* with that result.
+In mathematics, function composition is defined like this: ![ (f . g)(x) = f(g(x))](assets/images/higher-order-functions/composition.png), meaning that composing two functions produces a new function that, when called with a parameter, say *x*, is the equivalent of calling *g* with the parameter *x* and then calling the *f* with that result.
```
ulysses4ever: ["say" has to stay in commas](https://english.stackexchange.com/a/67733/155388), so the result would be: 

> ... two functions produces a new function that, when called with a parameter, say, *x*, is the equivalent of calling...

source_md/input-and-output.md
```diff
@@ -2149,7 +2149,7 @@ So the final result is an I/O action that will either act the same as the first
 ![non sequitor](assets/images/input-and-output/puppy.png){.right width=334 height=240}
 
 If you're familiar with *try-catch* blocks in languages like Java or Python, the `catch` function is similar to them.
-The first parameter is the thing to try, kind of like the stuff in the *try* block in other, imperative languages.
```
ulysses4ever, line 2152: This is very subtle but I think it was correct before. Because "other imperative language" would imply that Haskell is an imperative language. While "other, imperative" means that some other language, which is imperative in nature, would do this

ulysses4ever (commented):
Terrific, thanks! Some comments inline ..

source_md/input-and-output.md
```diff
@@ -478,7 +478,7 @@ Try playing with this program to get a feel for it!
 
 The `when`{.label .function} function is found in `Control.Monad` (to get access to it, do `import Control.Monad`).
 It's interesting because in a `do` block it looks like a control flow statement, but it's actually a normal function.
-It takes a boolean value and an I/O action if that boolean value is `True`, it returns the same I/O action that we supplied to it.
+It takes a boolean value and an I/O action. If that boolean value is `True`, it returns the same I/O action that we supplied to it.
```
MatthijsBlom: One sentence per line:
```suggestion
It takes a boolean value and an I/O action.
If that boolean value is `True`, it returns the same I/O action that we supplied to it.
```

source_md/input-and-output.md
MatthijsBlom: Well, it _is_ said that Haskell is the world’s finest imperative programming language 😏

But I agree that the comma should stay. The book itself, as a whole, does not consider Haskell an imperative language.

source_md/higher-order-functions.md
MatthijsBlom: I agree the second comma should stay, and don't feel adding a third is an improvement.

ulysses4ever: The when construct is parenthetical here (you can remove it and the sentence still makes the same point), so the new comma should be there for sure.

MatthijsBlom: I meant the original second comma, so I think I agree with you.
```
... a parameter, say, *x*, is ...
               1    2    3
```

ulysses4ever: Oh, I like the indexing! My point is that all three commas as you show above are necessary.

MatthijsBlom: The `, *x*,` bit wouldn't be parenthetical: without it the sentence would break, as $x$ is used later on.
Do you mean to pair the new, third comma with the one in `that, when`? If so, that pair wouldn't be parenthetical either, for the same reason.

I share some (presumed) urge to encode speech patterns in the writing, but also feel that this should be resisted. The purpose of comma's (and interpunction in general) is not so much to aid pronunciation as it is to aid reading, i.e. understanding.

How about

```suggestion
In mathematics, function composition is defined like this: ![ (f . g)(x) = f(g(x))](assets/images/higher-order-functions/composition.png), meaning that composing two functions produces a new function that, when called with a parameter, let's say *x*, is the equivalent of calling *g* with the parameter *x* and then calling the *f* with that result.
```

talestonini: I would agree "say" has to be inside commas in a body of strict language text. The tone of the book is more informal, so I believe we could drop that one for fluidity.

We still need the new (3rd) comma, because the "when" sentence is parenthetical. Parenthetical pertains to **syntax**, meaning syntactically the "when" sentence could be removed. The fact that the "x" introduced in the parenthetical sentence is referred to later pertains to **semantics**, so it does not invalidate the parenthetical classification.

In saying all that, I appreciate the spirit of correctness in this piece of review, and would settle with the suggestion introducing "let's" above :)

source_md/higher-order-functions.md
talestonini: No worries, I'll revert that.

source_md/input-and-output.md
talestonini: No worries, I'll revert that.

source_md/input-and-output.md
```diff
@@ -862,7 +863,7 @@ not a palindrome
 Even though we made a program that transforms one big string of input into another, it acts like we made a program that does it line by line.
 That's because Haskell is lazy and it wants to print the first line of the result string, but it can't because it doesn't have the first line of the input yet.
 So as soon as we give it the first line of input, it prints the first line of the output.
-We get out of the program by issuing an end-of-line character.
+We get out of the program by issuing an end-of-file character.
```
MatthijsBlom, line 866: Apparently GHCi plays weirdly with EOFs. We may want to note that somewhere. I'd like to understand _why_, too.

( Out of scope for this PR; just an observation. )

MatthijsBlom (approved):

talestonini:
Hi @ulysses4ever , I believe this is ready to merge ;)

ulysses4ever:
Hey @talestonini ! I'll double-check and hopefully merge by the end of week

ulysses4ever (approved):
Thanks a lot! Merging...

talestonini:
Thank you, @ulysses4ever !

PR #162, opened 2026-05-06
`let` in GHCi

MatthijsBlom:
This PR's main contribution is a note about `let` not being needed anymore when binding to names in GHCi.

I haven't touched the many `ghci> let` occurrences in the book. Some were already updated (#24), but keeping some or even many of the rest doesn't seem harmful. If it is desired that they are all removed, I'd still like to add a note that both forms exist and are synonymous, so that seeing `let`s in other older resources will be less surprising.

### DIFF
--- a/source_md/syntax-in-functions.md
+++ b/source_md/syntax-in-functions.md
@@ -583,9 +583,6 @@ ghci> boot
 Back in the day, using `let` used to be required when binding in GHCi.
 A bit inconvenient!
 These days you can omit the `let` and enjoy exactly the same result.
-But why, you may wonder, require `let` in the first place?
-The answer is that the notation at GHCi's prompt was modeled after a syntactic form called *`do`-notation*.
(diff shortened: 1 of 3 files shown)


source_md/syntax-in-functions.md
```diff
@@ -577,6 +577,17 @@ ghci> boot
 <interactive>:1:0: Not in scope: `boot'
 ```
 
+::: {.hintbox}
+**Let's not.**
+
+Back in the day, using `let` used to be required when binding in GHCi.
+A bit inconvenient!
+These days you can omit the `let` and enjoy exactly the same result.
+But why, you may wonder, require `let` in the first place?
```
ulysses4ever: I think it's going too much in the weeds. On top of that, I'd avoid forward references at all costs.

ulysses4ever (approved):
Thanks! And sorry for the delay!

PR #165, opened 2026-06-30
fix(types-and-typeclasses): Reword for clarity

CrownedPhoenix:
Nothing fancy - just came across a line that felt a little unclear.

### DIFF
--- a/source_md/types-and-typeclasses.md
+++ b/source_md/types-and-typeclasses.md
@@ -373,7 +373,7 @@ False
 `minBound` and `maxBound` are interesting because they have a type of `(Bounded a) => a`.
 In a sense they are polymorphic constants.
 
-All tuples are also part of `Bounded` if the components are also in it.
+All tuples are also part of `Bounded` if their components are part of `Bounded` as well.
 
 ```{.haskell: .ghci}
 ghci> maxBound :: (Bool, Int, Char)


source_md/types-and-typeclasses.md
```diff
@@ -373,7 +373,7 @@ False
 `minBound` and `maxBound` are interesting because they have a type of `(Bounded a) => a`.
 In a sense they are polymorphic constants.
 
-All tuples are also part of `Bounded` if the components are also in it.
+All tuples are also part of `Bounded` if the components within it are part of `Bounded`.
```
ulysses4ever: "components within" doesn't sound natural to my ear. How about

```suggestion
All tuples are also part of `Bounded` if its components are part of `Bounded` as well.
```

pierluc-codes: Alternative: All tuple types are instances of `Bounded` if all of their component types are instances of Bounded.

MatthijsBlom: Some day, I'd like to improve the book's language around class instances. As it stands today, the book is littered with phrases like "part of <_class_>" and "member of the <_class_> class". That is not how type classes are usually talked about, and also suggests that type classes are closed (i.e. one could make a list of all instances, or establish that something is _not_ an instance).

For now, I prefer @ulysses4ever's suggestion for its simplicity and consistency with the rest of the book.
It should say _their_ instead of _its_ though: _all tuples_ is plural.
Changing _tuples_ &mapsto; _tuple types_ would be more correct, but the former is probably already understood well enough.

CrownedPhoenix: Great - I've made the adjustments.

MatthijsBlom (approved):

pierluc-codes (approved):

ulysses4ever (approved):
Thank you!

PR #166, opened 2026-07-02
fix: operator is not a specifier for operator fixity

CrownedPhoenix:
I'm admittedly a noob to this but I'm extrapolating from 
```
`+`'s fixity is `infixl 6`
``` 
that the operator is not part of the fixity declaration. More specifically, the specification of the operator is the operand of the fixity declaration.

If that extrapolation is correct, then it is incorrect to say: 
```
`*`'s fixity is `infixl 7 *`
```
Since that would be like saying (in english)  
```
'*'s fixity is '*'s fixity is left-associative 7
``` 
as opposed to 
```
'*'s fixity is left-associative 7
```

### DIFF
--- a/source_md/making-our-own-types-and-typeclasses.md
+++ b/source_md/making-our-own-types-and-typeclasses.md
@@ -1040,7 +1040,7 @@ data List a = Empty | a :-: (List a) deriving (Show, Read, Eq, Ord)
 First off, we notice a new syntactic construct, the fixity declarations.
 When we define functions as operators, we can use that to give them a fixity (but we don't have to).
 A fixity states how tightly the operator binds and whether it's left-associative or right-associative.
-For instance, `*`'s fixity is `infixl 7 *` and `+`'s fixity is `infixl 6`.
+For instance, `*`'s fixity is `infixl 7` and `+`'s fixity is `infixl 6`.
 That means that they're both left-associative (`4 * 3 * 2` is `(4 * 3) * 2`) but `*` binds tighter than `+`, because it has a greater fixity, so `5 * 4 + 3` is `(5 * 4) + 3`.
 
 Otherwise, we just wrote `a :-: (List a)` instead of `Cons a (List a)`.


MatthijsBlom (approved):

MatthijsBlom:
The book's language is very loose. Don't read too much into it.

> I'm extrapolating from [...] that the operator is not part of the fixity declaration.

An operator's fixity declaration is the line of code, i.e. _a piece of text_, that defines how the operator should be read. This piece of text includes the operator's name. If it didn't, it wouldn't be clear which operator the fixity declaration is about.

Fixity declarations and fixities are different things! The former are pieces of text, the latter abstract properties.
When you are drinking a glass of water, you are not drinking the phrase "a glass of water"! No, you're drinking water.

> If that extrapolation is correct, then it is incorrect to say: [...]

It hardly matters. Typically people use natural language: `*` is left-associative and has precedence 7. When one writes `infixl 7` instead, that's really just a shorthand. There is no formality here.

`infixl 7 *` could serve as shorthand just as well. It wouldn't be wrong or anything. I just happen to prefer the other one.

> Since that would be like saying (in english) [...]

Computer code isn't meant to be a [cipher](https://en.wikipedia.org/wiki/Substitution_cipher): it needn't be transliterable. Also, it's totally OK to [borrow](https://en.wikipedia.org/wiki/Borrowing_(linguistics)) or to [code-switch](https://en.wikipedia.org/wiki/Code-switching). The purpose of language is to be understood and to understand others, not be be 'correct'.

CrownedPhoenix:
I'm so sorry - I totally did not mean to make a statement about the "correct" way to phrase this line. Though I absolutely concede that my earlier comment was poorly phrased in that regard. I completely agree with everything you've said.

When I read this line, I thought that it should either be:
```
For instance, `*`'s fixity is `infixl 7 *` and `+`'s fixity is `infixl 6 *`.
```
or 
```
For instance, `*`'s fixity is `infixl 7` and `+`'s fixity is `infixl 6`.
```

In other words, it just seemed like the two statements about operator fixity should match (in regard to whether they included the operator itself in the back-ticked text). It seemed to me that dropping `*` made the most sense - but either one would totally be fine.

Again - I'm really sorry that I came across as being pedantic. I'm just reading through this book and trying to leave it better than I found it. I have basically no opinions on what the "right" way it should read - just volunteering what felt more clear to me.

MatthijsBlom:
> I'm so sorry

No need!
I did not read you as being pedantic, or anything like that.

I share your feeling about consistency in the text, and have approved your change for that reason.

My comment was not intended as commentary on your change, but rather to clear up some confusion that I got the impression you might have. It was meant purely explanatory, but turned out a bit curt. Sorry about that.

ulysses4ever (approved):
This is a simple and straightforward fix. Thank you for taking the time to submit it! 

I'm sorry if the discussion here confused you a bit: we sometimes tend to overshare our vision of global issues I feel like :-)

PR #168, opened 2026-09-15
Minor fix (types-and-typeclasses): missing word

liakhovitch:
Added a missing word in 'Making Our Own Types And Typeclasses'.
My very first pull request to a public repo :)


source_md/making-our-own-types-and-typeclasses.md
```diff
@@ -882,7 +882,7 @@ data Either a b = Left a | Right b deriving (Eq, Ord, Read, Show)
 
 It has two value constructors.
 If the `Left` is used, then its contents are of type `a` and if `Right` is used, then its contents are of type `b`.
-So we can use this type to encapsulate a value of one type or another and then when we get a value of type `Either a b`, we usually pattern match on both `Left` and `Right` and we different stuff based on which one of them it was.
+So we can use this type to encapsulate a value of one type or another and then when we get a value of type `Either a b`, we usually pattern match on both `Left` and `Right` and we do different stuff based on which one of them it was.
```
ulysses4ever:
```suggestion
So we can use this type to encapsulate a value of one type or another.
Then, when we get a value of type `Either a b`, we usually pattern match on both `Left` and `Right`, and we do different stuff based on which one of them it was.
```

ulysses4ever (approved):
Thanks!

ulysses4ever:
I took the liberty to apply some more surgery on this arguably lengthy sentence: I split it into two and added more commas. As a result, it seems your authorship vanished. This wasn't intended, but it's probably hard to fix now. Your contribution is appreciated nevertheless!

